PR #81, opened 2020-07-30
add instructions to setup local-to-arcade transfers

nbrunett:


brianmajor (approved):
Thanks Nathan.  These will be need to be updated when the vos tools with cavern support are released to pypi, but this is useful to have around now so I'll merge it to the master branch.

PR #91, opened 2020-08-12
Added rsync and ssfhs instructions to README

drtobybrown:
Can you confirm that the new instructions are correct. I believe they work on MacOSX at least.


brianmajor (approved):
Looks correct to me.

PR #101, opened 2020-08-24
Add transfer verification suggestions

nbrunett:


brianmajor (changes requested):
File verification by size is not a reliable way see if two files are the same.  This is especially true in astronomy as different FITS files from a given instrument are often all the same size due to the nature of the FITS format.  So, I'd suggest either strengthening your warning or removing the size comparison method all together.

The printf syntax is slightly different on Macs but you've essentially made the point that.

Nice to have this type of documentation, thanks.

nbrunett:
I didn't know that about FITS files... sounds like some interesting Saturday morning reading.

I have removed the verification by size. I'm not sure what you meant about the Mac printf syntax, but I can tweak some more if needed.

PR #114, opened 2020-09-18
make locales consistent for directory verification with sort

nbrunett:
This actually caused me some confusion comparing directories on McMaster's astro machine.

I'm not an expert on shell variables and `export` so a quick sanity check around that would be appreciated. Thank you.


brianmajor (approved):
I'm not familiar with that either so I'll trust that you tried it out at least :)

PR #115, opened 2020-09-22
ct-156 ct-213 Updated software-contaiers/README.md

yeunga:
Please let me know if this file needs to be added to the Harbour repo.

### DIFF
--- a/software-containers/README.md
+++ b/software-containers/README.md
@@ -23,9 +23,6 @@ Use ENTRYPOINT to specify an executable to be run, for example a startup script 
 ### CMD
 Use CMD to specify a command to be executed at the end of the Dockerfile, "xterm".
 
-### desktop
-Each container is associated with a desktop file, for example terminal.desktop, which includes the path to the shell script, for example terminal.sh, which initializes and starts up the container.
-
 ### shell script
 A shell script is created for each container. The shell script initializes and starts up the container. As a result, PostAction is invoked. PostAction uses a default parameter to start up an xterm. However the shell script can pass a parameter to PostAction, which replaces the default parameter. The following is an example script that passes a parameter to PostAction:
 ```
--- a/software-containers/README.md
+++ b/software-containers/README.md
@@ -16,3 +16,18 @@ group:      sss files
 
 ### xterm
 `xterm` must be installed on the container
+
+### ENTRYPOINT
+Use ENTRYPOINT to specify an executable to be run, for example a startup script which can execute several tasks when the container starts up.
+
+### CMD
+Use CMD to specify a command to be executed at the end of the Dockerfile, "xterm".
+
+### desktop
+Each container is associated with a desktop file, for example terminal.desktop, which includes the path to the shell script, for example terminal.sh, which initializes and starts up the container.
+
+### shell script
+A shell script is created for each container. The shell script initializes and starts up the container. As a result, PostAction is invoked. PostAction uses a default parameter to start up an xterm. However the shell script can pass a parameter to PostAction, which replaces the default parameter. The following is an example script that passes a parameter to PostAction:
+```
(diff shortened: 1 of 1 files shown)


software-containers/README.md
```diff
@@ -16,3 +16,18 @@ group:      sss files
 
 ### xterm
 `xterm` must be installed on the container
+
+### ENTRYPOINT
+Use ENTRYPOINT to specify an executable to be run, for example a startup script which can execute several tasks when the container starts up.
+
+### CMD
+Use CMD to specify a command to be executed at the end of the Dockerfile, "xterm".
```
brianmajor, line 24: What happens if both ENTRYPOINT and CMD are provided?  What does that parameter mean in each scenario?  These types of questions should be answered.  I think maybe a table or matrix with all the combinations and results might be the best way to capture this.

software-containers/README.md
```diff
@@ -16,3 +16,18 @@ group:      sss files
...
+
+### desktop
+Each container is associated with a desktop file, for example terminal.desktop, which includes the path to the shell script, for example terminal.sh, which initializes and starts up the container.
+
```
brianmajor: This 'desktop' section is part of creating the menu for ARCADE.  Users won't do this so please remove.

software-containers/README.md
```diff
@@ -16,3 +16,18 @@ group:      sss files
...
+### shell script
+A shell script is created for each container. The shell script initializes and starts up the container. As a result, PostAction is invoked. PostAction uses a default parameter to start up an xterm. However the shell script can pass a parameter to PostAction, which replaces the default parameter. The following is an example script that passes a parameter to PostAction:
+```
```
brianmajor, line 28: A shell script isn't required for a container to run in arcade.  It is a good option to document though.

"A shell script is created" makes it sound like it is automated somehow.  It would be the users creating such a script.

software-containers/README.md
```diff
@@ -16,3 +16,18 @@ group:      sss files
...
+session-containers/arcade-desktop/software-scripts/arcade-feedback.sh:5:/opt/shibboleth/bin/curl -v -L -k -E /cavern/home/${arcade_username}/.ssl/cadcproxy.pem -d "target-ip=${IP_ADDRESS}" -d "software=canfar-registry.cloud.computecanada.ca/arcade-firefox:latest" --data-urlencode "param=https://github.com/opencadc/arcade/issues" https://${arcade_hostname}/arcade/session/${VNC_PW}/app
+```
```
brianmajor, line 30: Again, this is part of the installation so does not concern users.

brianmajor (changes requested):

brianmajor:
We also need to ensure that no matter how the container is built and whether or not a parameter is sent in that the title of the container is the key in arcade-software.properties.  You'll probably have to go through the code to see if that's true.

brianmajor:
Also, in jira, the phrase "and persist it in harbor" means document how users persist their containers in harbor (not persisting README.md).  We'll add that part once the harbor A&A is all cleared up.

software-containers/README.md
yeunga: Yes, the curl command can be executed manually.

Creation of a shell script is automated in ARCADE via the create-software-links.sh invoked within Makefile, no? Although if we want to port this README.md file to another project, this may not be true.

software-containers/README.md
yeunga: Okay, will delete it.

software-containers/README.md
yeunga: Is this README.md file for a user's consumption or for someone like a developer who builds a container?

brianmajor:
README.md here should be an instruction manual for arcade/petuan users on how to build a container.  It will also detail how to push images to harbor, but that can come in a second pull request.  So, it should not mention anything about how that container is installed or launched from arcade/petuan.

- Requirements for software in the container
- Requirements for configuration in the container
- How the different combinations of CMD and ENTRYPOINT will work with the defaults of arcade.  What to expect with the different combinations and if they are missing?
- Should users provide a wrapper startup script?  How do they ensure that script is executed?

Issue #58, opened 2020-05-15
GroupPermissions in Cavern via Filesystem


brianmajor:
When a POSIX file or directory is group read or write protected the users who are members of that group remain unauthorized when accessing the file or directory though the filesystem.

Group authorization works correctly when interacting through VOSpace.

SSSD is not considering all of the user's group memberships, only their primary group membership.

SSSD seems to be querying LDAP correctly for all group memberships.  The SSSD cache confirms this. (in /var/lib/sss/pipes/db)

----

Update: We haven’t been successful in having the root container process obtain the capability needed to attach the supplemental group memberships.  From experience in the k8s community, it seems the better approach is to set these groups explicitly set the securityContext info in the launch configuration.  SSSD is still needed to map names to the IDs.

If group membership changes during the lifetime of a pod, that change won’t be reflected in that pod.  However, this is the same as with the other approach, as memberships are attached to the main thread and never renewed, only inherited by other threads.

brianmajor:
This could also be caused by the unordered initialization sequence we are currently using in pods.   SSSD is being moved to a daemon set.  Another test should be performed after that change is made.

brianmajor:
Moving SSSD to a daemon set did not resolve this issue.

brianmajor:
The problem is fairly well understood now.  Here are some details.

Only supplementary groups are not showing up.  When the container is started as root, the supplementary groups show up.

When a process is started, the groups (primary and supplementary) are retrieved at that time and are attached to the process.  You can see these groups with `ps -eo pid,user,args,sgroup,supgrp` (supgrp is the list of supplementary groups, the ones missing when not root).

Running this command in both versions of the container shows that the one started as root have the supgrps, the other does not.

Because the initial process is started as root, the sub-processes just inherit the groups from the root process, so that explains why the processes owned by majorb still show the groups.

So, when looking at strace, the getGroups() system call just check the groups associated with the process.

The real work happens in a system call called initgroups(), which is called on the start of every process.  We only care about the root process because that will pass down the groups to the subprocesses.  So, it’s clear to me that root has the ability to get supplementary groups on initgroups(), but other users do not.

initgroups() itself seems to use other system calls to get those supplementary groups.  Specifically, getgroups() and setgroups().  I believe it is these two system calls that non-root users are unable to execute.

[...]

PR #119, opened 2020-09-29
Cadc 14 Sample test container to test the fix for issue #58

yeunga:
The alingas-terminal is a container created to test the solution to address issue #58. Main points in the solution.
- no ENTRYPOINT or CMD in Dockerfile
- the launch yaml file (in this case launch-alingas-terminalyaml) invokes a generic launch-software.sh script passing all runtime variables that may be useful to a user built container (in this case alingas-terminal)
- launch-software.sh runs arcade-init.sh script, if provided by the user, as the user
- arcade-init.sh must not block
- launch-software.sh executes xterm (blocking)


brianmajor (changes requested):
If you type `env` in a container, you will see that nearly all the parameters you have listed are already available.  Please reduce them to just the two that are required right now:  $CONTAINER_NAME (should be called $IMAGE_ALIAS) and $USER_NAME (should be $USERNAME)

software-containers/alingas-terminal/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apt-get update && yes | unminimize 
+RUN apt-get update && yes | apt-get install wget \
+    build-essential xterm vim rsync mlocate
+
+SHELL ["/bin/bash", "-c"]
+
+RUN apt-get update && yes | apt-get install sssd libnss-sss libpam-sss
+
+RUN mkdir /arcade
+ADD launch-software.sh /arcade
```
brianmajor, line 13: The logic in the launch software script should not be in the container itself

software-containers/alingas-terminal/arcade-init.sh
```diff
@@ -0,0 +1,37 @@
...
+        then
+            echo "Failed to execute $cmd, return code = $rc, $msg"
+        else
+            echo "Failed to execute $cmd , return code = $rc, no stdout or stderr"
+        fi
+    fi
+}
+
+#echo "starting xterm $1"
+#execute xterm "-fg white -bg black -title $1"
```
brianmajor, line 37: From where is execute() called?

brianmajor (commented):
I'm going to close this pull request soon as this test container of course will not be in git.

PR #121, opened 2020-10-06
add info on how (Ana)conda is setup on ARCADE

nbrunett:
This is pieced together from me poking around a bit, so my description may not be perfectly accurate.

Before this is merged, @drtobybrown should check that he can make his own environments that are stored in his home directory, and can access them by name by just following the conda docs that I linked to. If there are issues, then I may be able to add steps to this that will help.

### DIFF
--- a/arcade-tutorial/README.md
+++ b/arcade-tutorial/README.md
@@ -126,6 +126,20 @@ Finally, running this on your local machine and ARCADE will put the locale setti
 
 `export LC_ALL=$OLD_LC_ALL`
 
+# How to manage Python environments with conda on ARCADE
+
+Installing and managing the versions of your Python packages can be done with conda (e.g. working with a set of Python packages that you can export as a text list for others to exac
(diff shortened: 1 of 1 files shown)


drtobybrown:
Just confirming that conda is working fine for me on ARCADE. I was able to follow the steps on the getting started page for conda. 

I couldn't add packages to astroconda (permissions issue below), however, I was able to add packages to my own environment. This means that users may have to start from scratch for getting their conda set up how they like it.

```
EnvironmentNotWritableError: The current user does not have write permissions to the target environment.                              
  environment location: /opt/anaconda/envs/astroconda                                                                                 
  uid: 184130476                                                                                                                      
  gid: 184130476  
```

nbrunett:
I think the system-wide astroconda should be a "read-only" environment since everyone can access it. But there should be ways to duplicate that env to your own as a starting point that you could then modify (e.g. "clone the env" by exporting the environment packages to a .yml file and then creating a new env from that file).

Sounds like this is ready to merge from my point of view.

brianmajor (approved):

PR #132, opened 2020-12-08
CADC-8819 - Refactor skaha-wsproxy to remove duplicate code.

SharonGoliath:
Add some unit tests - 83% coverage.

### DIFF
--- a/skaha-wsproxy/conf/skaha_rewrite.py
+++ b/skaha-wsproxy/conf/skaha_rewrite.py
@@ -94,7 +94,7 @@ class Rewrite(object):
         self._cache = TTLCache(maxsize=max_size, ttl=ttl)
 
     def _build_url(self, segs, path, session_id, ip_address, params):
-        return None
+        raise NotImplementedError
 
     def _get_ip_for_session(self, session_id):
         session_ip_address = self._cache.get(session_id)
@@ -114,7 +114,7 @@ class Rewrite(object):
                            '-o',
                            'custom-columns=IPADDR:.status.podIP,'
                            'DT:.metadata.deletionTimestamp']
-                command_string = ' '.join([str(elem) for elem in command])
+                command_string = ' '.join(str(elem) for elem in command)
                 self.log('DEBUG: kubectl command: {}'.format(command_string))
                 command_output = subprocess.check_output(
                     command, stderr=subprocess.STDOUT)
--- a/skaha-wsproxy/conf/skaha_rewrite.py
+++ b/skaha-wsproxy/conf/skaha_rewrite.py
@@ -90,14 +90,7 @@ class Rewrite(object):
         :param ttl: time to live of cache entries
         """
         self._log_fqn = log_fqn
-        if os.path.exists(log_fqn):
-            self._log_file = open(log_fqn, 'a')
-        else:
-            dir_name = os.path.dirname(log_fqn)
-            if not os.path.exists(dir_name):
-                os.mkdir(dir_name)
-            self._log_file = open(log_fqn, 'w')
-
+        self._log_file = open(log_fqn, 'a')
         self._cache = TTLCache(maxsize=max_size, ttl=ttl)
 
     def _build_url(self, segs, path, session_id, ip_address, params):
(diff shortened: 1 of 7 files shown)


skaha-wsproxy/conf/skaha_rewrite.py
```diff
@@ -0,0 +1,183 @@
...
+        :param log_fqn: Fully-qualified name for the log file.
+        :param max_size: maximum size of the cache
+        :param ttl: time to live of cache entries
+        """
+        self._log_fqn = log_fqn
+        self._log_file = open(log_fqn, 'a')
+        self._cache = TTLCache(maxsize=max_size, ttl=ttl)
+
+    def _build_url(self, segs, path, session_id, ip_address, params):
+        return None
```
andamian: If this is meant to be abstract shouldn't raise a notimplementederror here?

skaha-wsproxy/conf/skaha_rewrite.py
```diff
@@ -0,0 +1,183 @@
...
+                           '--kubeconfig=/root/kube/k8s-config',
+                           'get',
+                           'pod',
+                           '--selector=canfar-net-sessionID={}'.format(
+                               session_id),
+                           '--no-headers=true',
+                           '-o',
+                           'custom-columns=IPADDR:.status.podIP,'
+                           'DT:.metadata.deletionTimestamp']
+                command_string = ' '.join([str(elem) for elem in command])
```
andamian: Simple `' '.join(command)` should work instead, no? All the elements seem to be str already. Why is command an array if a str is needed at the end?

skaha-wsproxy/conf/skaha_rewrite.py
```diff
@@ -0,0 +1,183 @@
...
+                sys.stdout.write(response + '\n')
+            else:
+                self.log('INFO: End response: None')
+                sys.stdout.write('https://www.canfar.net/notfound.html\n')
+        except Exception as e:
+            tb = traceback.format_exc()
+            self.log('ERROR: unexpected: {}:{}'.format(str(e), tb))
+            sys.stdout.write('https://www.canfar.net/notfound.html\n')
+        except:
+            self.log('ERROR: unclassified error')
```
andamian, line 177: I'm not sure I understand the reason this is done this way. I would catch `Error` above which covers everything. With this implementation, errors like simple syntax errors, will be tricky to detect.

skaha-wsproxy/conf/skaha_rewrite.py
```diff
@@ -0,0 +1,183 @@
...
+            sys.stdout.write('https://www.canfar.net/notfound.html\n')
+        sys.stdout.flush()
+
+    def log(self, message):
+        self._log_file.write(time.ctime() + u' - ' + message.encode() + u'\n')
```
andamian, line 182: Could use `logging` maybe?

andamian (approved):

skaha-wsproxy/conf/skaha_rewrite.py
SharonGoliath: It should, and now it does.

skaha-wsproxy/conf/skaha_rewrite.py
SharonGoliath: Removed the array.

skaha-wsproxy/conf/skaha_rewrite.py
SharonGoliath: Maybe. I don't know the logic behind what messages went to the log file, and what messages went to stdout, and I don't understand `logging` well enough to divide messaging that way without more time to research than I've got left on the task. I will if you think it's important though.

skaha-wsproxy/conf/skaha_rewrite.py
SharonGoliath: The effect is to not print a stacktrace for KeyboardInterrupt. I don't if that was on purpose or not, though. I can check with Brian?

brianmajor:
Just wanted to note how this has unique logging requirement.  stdout, stderr and stdin are used to communicate with apache, so nothing should be going to stdout for logging or error reporting.

PR #177, opened 2021-05-06
Fixed wrong group variable

kcasteels:


brianmajor (approved):
Looks good, glad you found this bug, but I don't think it would have caused the desktop display failure?  This fix changes the last thing allocate does.  The user was probably already a member of the group when using beta. (CADC-9391)

PR #198, opened 2021-06-03
Deploy cavern service as /arc (arbutus-cavern) - CADC-9639

brianmajor:


brianmajor:
@drtobybrown This latest release changes the cavern endpoint in ws-uv.canfar.net from /cavern to /arc.  I don't think you reference this anywhere in the arcade documentation but I thought I'd give you a heads up anyhow.

PR #212, opened 2021-06-29
CADC-9761

kcasteels:
Permissions are set to 700 on new home directory creation


brianmajor (approved):
Looks good Kevin.  Thanks for changing 'bucket' to 'images'.

PR #230, opened 2021-10-20
Morphtool menu entry added

kcasteels:
A menu link has been added to launch the morphtool app.


brianmajor (changes requested):
I believe you need to add the image entry to the other properties file in the k8s config too.

PR #248, opened 2021-12-08
reorganized deployment focussed code

brianmajor:


kcasteels (approved):
Changes look good and simplify things for the better.

PR #250, opened 2021-12-15
Cadc 10154a Clean up dangling sessions

yeunga:
Added a clean up script to clean up dangling sessions. 
Also added a corresponding README file.


deployment/k8s-config/scripts/cleanup.sh
```diff
@@ -0,0 +1,35 @@
...
+      # check if there is a corresponding pod
+      out=`kubectl -n skaha-workload get pods | grep -e "$id"`
+      if [[ -z "$out" && -n "$id" ]]
+      then
+        # no corresponding pod, delete orphaned sessions
+        if [ "$type" = "service" ]; then
+          kubectl -n skaha-workload delete service $name
+        else
+          kubectl -n skaha-workload delete ingress skaha-${session}-ingress-$id
+        fi
```
brianmajor: Why isn't $name used for the ingress deletion?  If it were you could probably find a way to remove the 'if' statement.

brianmajor (changes requested):

deployment/k8s-config/scripts/cleanup.sh
yeunga: You are right. This was a left over from previous code when I was using the name from 'kw get service' to construct a name for the ingress session. No need to do that now. Good eye. Will update the code and test it.

brianmajor (approved):
Looks good

PR #273, opened 2022-02-09
Cadc 9027a Implement a feature to build the AstroSoftware menu and associated menu items dynamically

yeunga:
The Resolution menu items are also built dynamically at start up time of the skaha-desktop container.


containers/session-containers/skaha-desktop/VERSION
```diff
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor
 # build version tag: timestamp
-TAGS="0.4.13 $(date -u +"%Y%m%dT%H%M%S")"
+TAGS="0.4.14 $(date -u +"%Y%m%dT%H%M%S")"
```
brianmajor: Let's make this version 1.0.0

brianmajor (changes requested):
I think we should try to move _all_ of the menu building files out of the Docker build (out of session-containers/skaha-desktop) and into the runtime startup (skaha-workflow directory).

You can mount an entire directory as a config map.  That's where all the template files can go.  You should put them somewhere other than /dockerstartup (that will still be used for the NoVNC image).  

I'll have a closer look at the rest once that refactoring is done.

brianmajor:
There is a secondary goal of these changes:  to (mostly) remove skaha-specific modifications to the NoVNC container so that new versions of NoVNC (with xfce) can be brought in with minimal or no alteration and run in skaha.  So, that's why I'm recommending the custom menu files be moved to deployment configuration.  Also, they should be where the startup script is as there is a dependency there.

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
```diff
@@ -73,21 +73,9 @@ ADD ./src/common/scripts $STARTUPDIR
 RUN $INST_SCRIPTS/set_user_permission.sh $STARTUPDIR /headless
 
 # CANFAR Specifics Below
-
 COPY conf/xfce-applications.menu /etc/xdg/menus/xfce-applications.menu
 COPY conf/xfce-applications.menu /etc/xdg/menus/applications.menu
```
brianmajor, line 77: I think these two COPY lines (and the associated files) can be removed now?  The default one in the container should work?  (Maybe I'm wrong.)

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
```diff
@@ -115,16 +103,18 @@ RUN useradd -m -g guest -u 30001 guest && \
     ln -s /home/guest/.ssl /headless/.ssl
 
 # Add scripts that launch sofware containers
-ADD tmp/*.sh /usr/local/bin/
 ADD software-scripts/*.sh /usr/local/bin/
+RUN rm /usr/local/bin/dragons.sh
+RUN rm /usr/local/bin/nifty.sh
```
brianmajor: And remove this section?  (lines 105-108)

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
```diff
@@ -115,16 +103,18 @@ RUN useradd -m -g guest -u 30001 guest && \
...
 
 # Add desktop icon files
 RUN mkdir -p /headless/.icons
 COPY software-scripts/icons/* /headless/.icons/
 
 # Add desktop links to those scripts
 COPY software-scripts/*.desktop /usr/share/applications/
-COPY tmp/*.desktop /usr/share/applications/
+RUN rm /usr/share/applications/DRAGONS.desktop
+RUN rm /usr/share/applications/nifty.desktop
```
brianmajor: And remove this section?  (lines 114-117)

deployment/k8s-config/kustomize/base/skaha/config/launch-novnc.yaml
```diff
@@ -33,6 +33,9 @@ spec:
       hostname: "${software.hostname}"
       containers:
       - name: "${skaha.jobname}"
+        command: ["/skaha-system/start-desktop.sh"]
+	args:
+        - "uv.canfar.net"
```
brianmajor: This should be ws-uv.canfar.net

brianmajor (changes requested):

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
yeunga: No, the files and therefore the COPY lines are needed. They are merged with the AstroSoftware menu generated at deployment time.

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
yeunga: These .sh files are different from the generic ones. They are needed until we don't need these special .sh scripts.

yeunga: Will need to modify these two applications so that they can take the generic .sh arguments to start up.

Also these two apps are currently not in the list of desktop-app. To test the change, it's better to include them in the list.

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
yeunga: Yes, can remove these lines after the two apps have been updated to work with generic .sh.

deployment/k8s-config/kustomize/base/skaha/config/launch-novnc.yaml
yeunga: Updated. Thanks.

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
yeunga: There are .desktop files for other applications in software-scripts/., e.g. gedit.desktop, terminal.desktop. Unless we want to generate .desktop files for these apps, we need to keep line 115.

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
yeunga: Removed those two lines and the associated .sh files.

containers/session-containers/skaha-desktop/Dockerfile.centos.xfce.vnc
yeunga: Removed lines 116 and 117 and associated .desktop files.

brianmajor (approved):

PR #309, opened 2022-03-09
Forward the requested number of cores when launching Pluto

sefffal:
Continuing a slack conversation with @brianmajor ...

Julia processes, including Pluto notebooks, can easily make use of multiple threads if available.
But the normal way of detecting the number of cores (`julia -t auto`) always returns the number of physical cores on the node.
As an example, this leads to running with 32 threads when only 2 cores are provisioned, hurting performance.

If I understand Kubenetes, and I'm sure that I *don't*, this change will forward this information from Kubernetes to julia using the `software.requests.cores` and `JULIA_NUM_THREADS` environment variable. 

Unfortunately I don't have a way to test this change locally.
If this does work, I recommend we make a similar change to the other session containers. Most numerical software including Numpy (and probably Carta) falls back on BLAS libraries for matrix multiplication. To prevent the same oversubscription issue, we should set `OPENBLAS_NUM_THREADS` and `MKL_NUM_THREADS` to cover all our bases.

### DIFF
--- a/containers/session-containers/skaha-pluto/src/startup.sh
+++ b/containers/session-containers/skaha-pluto/src/startup.sh
@@ -20,7 +20,7 @@ export JULIA_LOAD_PATH="/opt/julia/environments/v1.7/Project.toml:"
 # Default project environment kept in the user's home directory
 export JULIA_PROJECT="$HOME"
 
-julia -t auto -J /PlutoSysImg.so -e 'using Pluto; Pluto.run(require_secret_for_access=false, auto_reload_from_file=true, launch_browser=false, host="0.0.0.0")'
+julia -J /PlutoSysImg.so -e 'using Pluto; Pluto.run(require_secret_for_access=false, auto_reload_from_file=true, launch_browser=false, host="0.0.0.0")'
 
 echo "Exiting"
 
--- a/deployment/k8s-config/kustomize/base/skaha/config/launch-notebook.yaml
+++ b/deployment/k8s-config/kustomize/base/skaha/config/launch-notebook.yaml
@@ -56,6 +56,14 @@ spec:
           value: "/arc/home/${skaha.userid}"
         - name: XDG_CACHE_HOME
           value: "/arc/home/${skaha.userid}"
+        - name: JULIA_NUM_THREADS
+          value: "${software.requests.cores}"
+        - name: OPENBLAS_NUM_THREADS
+          value: "${software.requests.cores}"
+        - name: MKL_NUM_THREADS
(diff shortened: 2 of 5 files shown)


brianmajor (changes requested):
Hi Will.  I have applied these changes to the launch yaml in skaha.  I have not published a new pluto image though.  If your recently pushed image `pluto-1.8.0-beta1` has the `-t auto` removed you should be able to test this now.

In this PR you please make the associated env changes to `launch-pluto.yaml` in ```deployment/k8s-config/kustomize/overlays/keel-dev/skaha/config/launch-pluto.yaml```

sefffal:
Great, thank you! I tested this and the thread counts are set correctly for Julia, OpenBLAS, and MKL:
image

IMO the latter two should probably be added to the general jupyter yaml file as well but other users might not agree.

I made the change in the other location as well. What is the difference between the two files?

brianmajor:
Okay, great!  The other file in the keel-dev folder is used in our development kubernetes environment.  They should be kept in sync, so thanks for doing that.  There are only small differences between them but unfortunately (due to the nature of the the kustomize tool we use) the entirety of the files need to exist in both trees.

I agree that that environment should be added to the jupyter sessions as well.  I can't see it doing any harm.  Do you want to make that change as part of this PR?  The two files (dev & prod) for jupyter are named `launch-notebook.yaml`

sefffal:
Ok sure I’ll go ahead and update Jupyter too. Looks like `OMP_NUM_THREADS` should also be set in that case.

What about CARTA? Does it know how many threads to use?

brianmajor:
Good question about CARTA.  I don't know, but on this page, if you search for the word `thread` there is some discussion of it:  https://github.com/CARTAvis/manual_source/blob/master/source/installation_and_configuration.rst

sefffal:
Done for `notebook`.

brianmajor:
For CARTA, there is a command line option for manually setting the thread count.  So it's possible to do this by:
- in launch-carta.yaml, add a 3rd arg to the command that is "software.requests.cores".
- in skaha-carta/src/skaha-carta: read that third param and make it the value of the -t argument (or whatever it is).

But, since there are multiple carta versions with different APIs, and since we have no proof that it isn't working properly already I suggest we leave those changes out of this PR.

sefffal:
Okay makes sense. In that case I think this is ready to go.

brianmajor (changes requested):
Sorry, one more change:  could you please increment the minor version number in session-containers/skaha-pluto?

Issue #140, opened 2021-01-29
notebooks: start dir is /home/jovyan, not $HOME


brianmajor:
Terminals in notebooks should open to $HOME (/cavern/home/<userid>)

brianmajor:
Looks like this is still an open Jupyter lab issue and is not yet possible:

https://github.com/jupyterlab/jupyterlab/issues/1366

chriswillott:
Yes, if I recall correctly what fixed it was adding a simple 'cd' to the default bashrc file when building the Docker image.

I have a file called global_bashrc that contains this (it also sets the prompt and activates the correct conda environment):
```
export PS1="${debian_chroot:+($debian_chroot)}\u@notebook:\w\$ "
# Enable conda and activate jwstdp by default
. "/opt/conda/etc/profile.d/conda.sh"
conda activate canucs0p5
cd

``` 

The Dockerfile has these lines:
```
# Add content to the global bashrc
COPY global_bashrc /home/jovyan
RUN cat /home/jovyan/global_bashrc >> /etc/bash.bashrc &&\
    rm /home/jovyan/global_bashrc

```

sfabbro:
The "official" workaround for container images based on `jupyter/scipy-notebook` is to create a [startup hook script](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#startup-hooks) and place it it `/usr/local/bin/start-notebook.d/` , e.g.
```
#!/bin/bash
cd $HOME
```
That way the terminal will start in the `/arc/home/<user>` directory.

BearBearCodes:
> The "official" workaround for container images based on `jupyter/scipy-notebook` is to create a [startup hook script](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#startup-hooks) and place it it `/usr/local/bin/start-notebook.d/` , e.g.
> 
> ```
> #!/bin/bash
> cd $HOME
> ```
> 
> That way the terminal will start in the `/arc/home/<user>` directory.

Hmm... I noticed you did this to get the `skaha-notebook:22.02` terminals to open in `/arc/home/<USER>`. Weirdly enough, if I copy your suggestion, my terminal starts in `/home/jovyan`.

sfabbro:
Make sure your hook script is executable. You can also check the session logs.

BearBearCodes:
[...]

PR #311, opened 2022-03-11
Starting in HOME directory

sfabbro:
Two trivial fixes for notebook users:
- possibility for users to add jupyterlab options at runtime by setting the JUPYTERLAB_ARGS environment variable, i.e. `ENV JUPYTERLAB_ARGS="--debug"`.
- workaround to ensure the notebook terminal starts in the /arc/home/<user> (issue #140)


brianmajor (dismissed):

brianmajor:
I'm going to test this before merging.

brianmajor:
A couple of things:
- Would it be better if the environment values were supplied at container launch time rather than at build time?  For headless sessions, environment can be passed in on launch with the `env=` param.  We could do that for the other sessions too.
- the cd $HOME trick seems to work
- The "$@" is passing the session ID to lupyter lab.  Jupyter lab thinks this is a ipynb file so fails.  If we go this way I think the "$@" has to be removed.

brianmajor:
I just realized that my comment about dynamic environment setting is really part of a different discussion.  The var in the scripts would have to be there anyhow.
Another option for passing args to jupyter is to use an `arg=` param on the launch which might be a better match.

sfabbro:
Yes doing it a runtime with an `env=` parameter would be even better (or both build and runtime). The UI would need to reflect this possibility too. Hopefully in this case users would not need to pass even more arguments through the "$@" item

brianmajor:
> The "$@" is passing the session ID to lupyter lab. Jupyter lab thinks this is a ipynb file so fails. If we go this way I think the "$@" has to be removed.

@sfabbro -- you need to remove the "$@" from the script.

sfabbro:
done

brianmajor:
Thanks.

I don't see file `skaha-hook.sh` in this PR so I think that needs to be added.  What does it do?

sfabbro:
skaha-hook.sh is not needed anymore. it used to do the "cd home".

brianmajor:
Ah, sorry, misread the diff.

brianmajor:
This is released now @sfabbro   Not sure if you built and pushed the new skaha notebook image, but since it only removed the hook there's no rush.

PR #313, opened 2022-03-30
Harbor backup and restore

kcasteels:


deployment/k8s-config/scripts/harbor-backup.sh
```diff
@@ -0,0 +1,54 @@
...
+#***************************************************************
+
+kubectl exec -it --namespace=cadc-harbor $POSTGRES_POD -- bash -c "export PGPASSWORD=$PGPASSWORD ; mkdir -p $PGDUMP_DIR ; pg_dump -U postgres registry > $PGDUMP_DIR/registry.back ; pg_dump -U postgres postgres > $PGDUMP_DIR/postgres.back ; pg_dump -U postgres notarysigner > $PGDUMP_DIR/notarysigner.back ; pg_dump -U postgres notaryserver > $PGDUMP_DIR/notaryserver.back"
+kubectl exec -it --namespace=cadc-harbor $POSTGRES_POD -- bash -c "export AWS_ACCESS_KEY_ID=$KEY_ID ; export AWS_SECRET_ACCESS_KEY=$ACCESS_KEY ; export RESTIC_PASSWORD=$PASSWORD ; ./tmp/restic_0121 -r $RESTIC_REPO --verbose backup $PGDUMP_DIR"
+
+#***************************************************************
+#Backup Redis
+#***************************************************************
+
+kubectl cp restic_0121 cadc-harbor/$REDIS_POD:/tmp/.
```
brianmajor, line 42: Should all these `kubectl cp` commands specify a namespace?

brianmajor, line 42: What and where is file `restic_0121`?

deployment/k8s-config/scripts/harbor-backup.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+#https://github.com/restic/restic/releases/download/v0.12.1/restic_0.12.1_linux_amd64.bz2
+#https://restic.readthedocs.io/en/stable/index.html
+#https://github.com/stonezdj/harbor/blob/backup_restore/tools/harbor-backup.sh
+
+RESTIC_REPO='s3:repo-url'
+KEY_ID='<aws access key id>'
+ACCESS_KEY='<aws access key>'
+PASSWORD='<restic password>'
```
brianmajor, line 10: Where are these credentials being stored?

brianmajor (changes requested):

brianmajor:
@kcasteels When it comes to making the production versions of the scripts, will there be differences?  If so should they be named according to their environment?  Perhaps they should go somewhere in the `harbor` subfolder as well.

deployment/k8s-config/scripts/harbor-backup.sh
kcasteels: The namespace is specified by "cadc-harbor/$REDIS_POD" where the first part is the namespace and the second part is the pod name.

kcasteels: I'll include the curl command to download the binary.

deployment/k8s-config/scripts/harbor-backup.sh
kcasteels: I have them stored locally on my work machine. Is there a better place to put these?

deployment/k8s-config/scripts/harbor-backup.sh
kcasteels: I commited the restic amd64 linux binary since restic is released as a bzip2 binary and that isn't installed on the keel nodes by default.

deployment/k8s-config/scripts/harbor-backup.sh
brianmajor: I don't know... is there an internal git repo that ops uses for secrets?

kcasteels: It seems we don't have a consistent system in place for secrets between ops people. Setting that up might be a story in itself. For now I have the secrets in an encrypted database stored locally.

PR #317, opened 2022-04-14
add k8s-config files for CASTOR ETC GUI

BearBearCodes:
The Dockerfile and other source code are in a private repository. I tried to emulate the `pluto` container type.

Also, originally I had the following line at the end of my Dockerfile:

```dockerfile
CMD ["gunicorn", "-b", "0.0.0.0:5000", "connector:app", "--log-level=debug", "--log-file=/dev/stdout"]
```

But since I'm not sure if the `command` entry in `launch-castor-etc.yaml` is required, I replaced the line above with the following code:

```dockerfile
ADD docker/startup.sh /skaha/
RUN chmod u+x /skaha/startup.sh
```

with `startup.sh` containing:
```bash
#!/bin/bash
LOGLEVEL=${1:-debug}
echo "Starting gunicorn..."
cd /backend
gunicorn -b 0.0.0.0:5000 connector:app --log-level=${LOGLEVEL} --log-file=/dev/stdout
```

### DIFF
--- a/deployment/k8s-config/kustomize/base/skaha/config/ingress-castor-etc.yaml
+++ b/deployment/k8s-config/kustomize/base/skaha/config/ingress-castor-etc.yaml
@@ -1,0 +1,28 @@
+# apiVersion: traefik.containo.us/v1alpha1
+# kind: Middleware
+# metadata:
+#   name: skaha-castor-etc-middleware-${skaha.sessionid}
+# spec:
+#   replacePathRegex:
+#     regex: ^/session/castor-etc/${skaha.sessionid}(/|$)(.*)
+#     replacement: /$2
+
+# ---
+
+apiVersion: traefik.containo.us/v1alpha1
+kind: IngressRoute
+metadata:
+  name: skaha-castor-etc-ingress-${skaha.sessionid}
+spec:
+  entryPoints:
+    - web
+  routes:
+  - kind: Rule
+    match: Host(`ws-uv.canfar.net`) && PathPrefix(`/session/castor-etc/${skaha.sessionid}/`)
+    services:
+    - kind: Service
+      name: skaha-castor-etc-svc-${skaha.sessionid}
+      port: 5000
+      scheme: http
+    # middlewares:
+    #   - name: skaha-castor-etc-middleware-${skaha.sessionid}
--- a/deployment/k8s-config/kustomize/base/skaha/config/launch-castor-etc.yaml
+++ b/deployment/k8s-config/kustomize/base/skaha/config/launch-castor-etc.yaml
@@ -40,6 +40,8 @@ spec:
           value: "${skaha.userid}"
         image: ${software.imageid}
         command: ["/skaha/startup.sh"]
+        args:
+        - debug  # verbosity level of logs
         imagePullPolicy: Always
         resources:
           requests:
--- a/deployment/k8s-config/kustomize/base/skaha/config/launch-castor-etc.yaml
+++ b/deployment/k8s-config/kustomize/base/skaha/config/launch-castor-etc.yaml
@@ -41,6 +41,7 @@ spec:
         image: ${software.imageid}
         command: ["/skaha/startup.sh"]
         args:
+        - ${skaha.sessionid}  # for URL routing
         - debug  # verbosity level of logs
         imagePullPolicy: Always
         resources:
--- a/deployment/k8s-config/kustomize/base/skaha/config/launch-castor-etc.yaml
+++ b/deployment/k8s-config/kustomize/base/skaha/config/launch-castor-etc.yaml
@@ -1,0 +1,84 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    canfar-net-sessionID: "${skaha.sessionid}"
+    canfar-net-sessionName: "${skaha.sessionname}"
+    canfar-net-sessionType: "${skaha.sessiontype}"
+    canfar-net-userid: "${skaha.userid}"
+  name: "${skaha.jobname}"
+spec:
+  activeDeadlineSeconds: 1209600
+  ttlSecondsAfterFinished: 86400
+  template:
+    metadata:
+      labels:
+        canfar-net-sessionID: "${skaha.sessionid}"
+        canfar-net-sessionName: "${skaha.sessionname}"
+        canfar-net-sessionType: "${skaha.sessiontype}"
+        canfar-net-userid: "${skaha.userid}"
+        job-name: "${skaha.jobname}"
+    spec:
+      restartPolicy: OnFailure
+      ${skaha.schedulegpu}
+      imagePullSecrets:
+      - name: ${software.imagesecret}
+      securityContext:
+        runAsUser: ${skaha.posixid}
+        runAsGroup: ${skaha.posixid}
+        fsGroup: ${skaha.posixid}
+        supplementalGroups: [${skaha.supgroups}]
+      priorityClassName: uber-user-preempt-medium
+      hostname: "${software.hostname}"
+      containers:
+      - name: "${skaha.jobname}"
+        env:
+        - name: skaha_hostname
+          value: "${skaha.hostname}"
+        - name: skaha_username
+          value: "${skaha.userid}"
+        image: ${software.imageid}
+        command: ["/skaha/startup.sh"]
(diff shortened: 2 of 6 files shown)


brianmajor:
Hi @BearBearCodes -- Looks like a good start.  I think we should aim to get this to a place where it can be tested in the platform.  We have a 'dev' system which I think would be appropriate for this.  Could you please add the same three files to the `kustomize/overlays/keel-dev/skaha/config` directory?  I think the only difference is that the hostname there is rc-uv.canfar.net (not ws-uv).  Cheers.

BearBearCodes:
Thanks @brianmajor! I've pushed the 3 files to the `kustomize/overlays/keel-dev/skaha/config` directory, where I've changed the single `ws-uv` instance to `rc-uv`.

BearBearCodes:
Hmm... After pushing my Dockerfile to rc-uv CANFAR and trying to run the image, I see the following error message when I look at the logs:

```bash
standard_init_linux.go:228: exec user process caused: permission denied
```

This happens whether I use `startup.sh` or `CMD ["gunicorn", ...]`. On my local machine, I've confirmed that both of these methods work to start the Flask server & GUI.

BearBearCodes:
Okay I think I figured out the problem. I had my user as `root` before, so using `u+x` did not actually change the permissions of `startup.sh` to be user-executable. For now, I have changed that command to `chmod a+x`.

Now I have to figure out how to handle the dynamic URL paths you were referring to, which I thought I had worked around (i.e., I cannot assume requests start at `/`). I wonder if there is a way that the Docker container can give the `sessionID` to Flask, as I wasn't able to figure out how to get Flask to recognize arbitrary connection paths...

BearBearCodes:
This new commit fixes the URL routing in Flask, but the static JS files from the frontend (created and minified via `npm run build`) are blind to this environment variable (they are built assuming the app is at the server root, see https://create-react-app.dev/docs/deployment/#building-for-relative-paths). One way to fix this is by building the app at runtime (!) and pass the `sessionID` to the build process, which would _severely_ slow down container launch times. Another might be to run `npm start` with the unminified source files, but this would lead to worse GUI performance and `npm start` is not meant for production anyway. Tricky... Any thoughts/suggestions?

BearBearCodes:
YES!!! I have the static files working now!!!

Here is a summary of what I changed (in my source code). Not sure if anything before the last item is actually necessary (I'll do more thorough testing another day).

1. Added the following code in `manifest.json`:

```json
  "start_url": "./",
  "scope": ".",
```

2. Added the following code in `package.json`:

```json
  "homepage": ".",
```

3. Changed the Flask app's `static_folder` to an absolute path, namely `static_folder="/backend/client"`. I highly doubt this was necessary in getting the app to work, but I'm not changing it now...

4. Used the following 2 regex searches to find & route static files. Note that these static files are being delivered safely by Flask (i.e., the user cannot manipulate any URLs to try and access any filesystems, not that I have any volumes mounted in the Docker container anyway).

```python
        after_url = re.search(r"(?<=session/castor-etc/).*$", path).group()
        filename = re.search(r"[^/?]*\.(?:gif|png|jpeg|jpg|ico|js|css)$", path)
```

brianmajor:
Hi Isaac.  Are you waiting for anything on our side?

BearBearCodes:
Hi @brianmajor, are you able to merge my most recent commit (https://github.com/opencadc/skaha/pull/317/commits/53275b3be31c9be49cd57c3af69cfbd2767be7c2) into rc-uv CANFAR to test (https://github.com/brianmajor/skaha/tree/castor_etc_gui)? Thanks again

brianmajor:
Okay, done.  You'll need to start new sessions to test of course.  We will have to think about where things belong a bit... the concept of this session type castor-etc exists in opencadc, but the startup script is in CASTOR-telescope, which creates an undesirable  inter-dependency.  We can look at that more after you get things working.

BearBearCodes:
Okay thank you so much. I'll test the new container sometime later today once I finish my current task (might take a while). We can definitely talk about the location of the startup script and other files later.

BearBearCodes:
Hi @brianmajor, I have the GUI working on CANFAR now! Would it be possible for you to:
1. merge this PR with the ws-uv CANFAR codebase, and 
2. change the `castor` project permissions on Harbor such that only Dr. Tyrone Woods and I can access it for the time being?

Or would you prefer to sort out the startup script inter-dependency first? In this case, we can probably chat about this sometime Friday (or whenever you're free). I say Friday only because the upcoming 2 days are very hectic for me.

brianmajor:
Hi @BearBearCodes 

I will make this session type available on ws-uv, though I don't want to add it as an option for users to pick this type just yet.  If the only available image of this type is private then users won't be able to run it so would cause confusion.  (The 'types' list is unfortunately built into the portal UI.). Though, once an image is launched (via curl) it will show up in the portal and can be connected to and deleted from there.  The other option is to not have the image be private.

In the longer term there are some things to be worked out:

1. As mentioned earlier, the decoupling of logic between the image and the platform needs to be sorted.
2. I'm hoping that your team plans on making the image public in the long term so that others can benefit from it.
3. If no to point 2, we may leave it as a 'hidden' feature for a while.
4. If yes to point 2, then we should think about how it can be maintained going forward.  The fact that the source is external at the moment could be problematic.  (And yes, I know I stated the opposite earlier, but hadn't thought it through.)
5. The platform does a good job of supporting private/proprietary images of type 'desktop-app' (those launched from within a desktop session) but not of the 'session' types (those which are available via a browser like castor-etc and notebooks).  This is mainly due to the list of types being built into the UI like I said above.
6. We should give consideration to making all these session types more common so that the platform doesn't have to know about them.  This would mean using the same 'yaml' k8s files for the different sessions, which, in turn, essentially means using the same port and path resolution technique.

I will turn points 5 and 6 into issues to be worked on at some point.

brianmajor (approved):

brianmajor:
@BearBearCodes Could you please test on ws-uv now?  I'm thinking that there's probably no need to make the image private since it's not a launch option in the portal?

BearBearCodes:
Thanks so much for your help @brianmajor. I can confirm the GUI works on ws-uv CANFAR. :)

Regarding your comments:
1. Yep we can definitely work this out. If it is convenient, we can set up a Zoom meeting with myself, Dr. Woods, and maybe Dr. Côté as well.
2. We will definitely make the image public in the (very short) future.
3. n/a
4. We can discuss this when we are figuring out item 1.
5. If the desktop app has docker installed, I can also launch the GUI from there (I haven't looked at the documentation for this yet). But I suspect we will strongly prefer to have this tool running as a session rather than as a desktop app on CANFAR.
6. Okay, this is something I will think about and relay to future CASTOR projects that may want to be hosted on CANFAR.

Thanks again for your help!

brianmajor:
2.  Okay, great.

5.  I haven't seen it yet but sounds like it is not suited for the desktop.

BearBearCodes:
Hi @brianmajor, I mentioned the interdependency issue to Dr. Woods and Dr. Côté and they agree that it makes sense to have the `startup.sh` script on this repo. Regarding the source code and the Docker container, though, is it at all possible to keep that in our `CASTOR-telescope` GitHub organization? The source code is only for running within the Docker container and nothing would be touching the outside environment—I did not mount any filesystems in the k8s config.

Summary:
- we will keep the startup.sh script on `opencadc/skaha`
- we want to keep the source code on `CASTOR-telescope`

brianmajor:
Hi @BearBearCodes .  I've given this some more thought.  If you think of the presence of file /skaha/startup.sh to be part of the deployment API of the platform, then I think we could consider that an acceptable decoupling of responsibility.  So, in short, I think it's fine to keep you source code in your own repo and also to keep the startup script there.  In the longer term we can think of using this paradigm with the other session types and converge on the launch pattern.

BearBearCodes:
Thanks for your reply @brianmajor. I will relay this information to the others to keep in mind for future projects as well. If any future software needs to launch a web app + server (i.e., via `gunicorn`), then we can definitely try to merge them under the same session type.

Thanks again for all your help getting this set up on CANFAR! :D

brianmajor:
Sounds good Isaac.  I think we'll move toward supporting a general session
type where the port is fixed (5000 is fine) and the expectation is that
images have a file /skaha/startup.sh that will receive the session ID as a
parameter.

PR #323, opened 2022-04-27
Introduce 'contributed' session type

brianmajor:
The intent of this PR is to allow users to develop and launch session (ie web) images without having to modify any of the platform code or deployment infrastructure.  Developers can build these applications knowing that:

1. Their instance must listen on port 5000
2. Their instance will receive requests at the root path (/), but URLs in the browser will look like https://<host>/<path>, where `<host>` and `<path>` are subject to change.  This path will initially be `https://ws-uv.canfar.net/sessions/contrib/<sessionid>` 
3. The instance must be started by a script in the image available at `/skaha/startup.sh` and will be passed 1 parameter, the sessionid.

@BearBearCodes We are going to move castor-etc to this contributed model.  I don't think there are many changes on your side, except:

- For point 2 above: because you removed the 'middleware' component of the ingress, I assume you have somewhere in your image a string like `"https://ws-uv.canfar.net/sessions/castor-etc" + <sessionid>` .  This shouldn't be necessary, as I understand that the URL in the browser can be determined programatically somehow.  @sefffal may know the details of this as that's how the pluto session (and the notebook and carta sessions) works.  If you want to defer this work till later you can just change `castor-etc` to `contrib` for now.

Once complete the 'contributed' session type will be added to the portal and the CASTOR UI image will be shown there.  (I'll update the label at that time.).  You may want to consider renaming the image to FORECASTOR (or whatever the users will be familiar with).

@sefffal Eventually we'd like to move pluto to this category as well, and maybe even all the sessions so that they all follow a common deployment API.

If either of you have comments or suggestions for improvement I'd be grateful to hear them.


BearBearCodes:
> This shouldn't be necessary, as I understand that the URL in the browser can be determined programatically somehow

Correct. I am just sending requests to whatever is the URL shown in the browser. But the backend must know where to listen for incoming requests, which is why I currently have a string with "session/castor-etc" hard-coded (only) in the backend.

> the 'middleware' component of the ingress

Ah I just realized I can restore the "middleware" component of the ingress, which looks like it will replace the string `/session/castor-etc/${skaha.sessionid}(/|$)(.*)` with just `/<path>`, where `<path>` is anything that comes after `/session/castor-etc/${skaha.sessionid}/`. Is this correct? This might actually simplify a lot of the process (and I would no longer require the `sessionID` parameter in `/skaha/startup.sh`).

BearBearCodes:
Actually, you _did_ restore the middleware component, so the frontend should send requests to `https://ws-uv.canfar.net/sessions/contrib/<sessionid>(/path?)` and the backend should listen for requests at `/<path>`, correct?

sefffal:
Hi @brianmajor  I think this is a great direction for the platform to move!

I have a couple suggestions.
First, I think as much as possible of the launched configuration should be exposed through standardized environment variables. We have some now for e.g. the number of cores that are specific to different software. These could become CADC_NUM_CORES for example.

Second, could we use this opportunity to standardize the launch process a bit? For example respecting the image CMD and ENTRYPOINT instead of using startup.sh. These could of course point to a startup script if desired.
Similarly I’d be interested if there’s a way to avoid copying those setup files into the image.

With those changes it would become much easier to adapt any docker image exposing a web service.

thoughts?


> URL can be determined programmatically somehow
Yes that’s right from JavaScript you can find this out and even modify it programmatically. 
Reverse proxies have been around for a long time so nowadays most well architected web apps can run out of subdirectories so it shouldn’t be an issue most of the time.

brianmajor:
> Actually, you did restore the middleware component, so the frontend should send requests to https://ws-uv.canfar.net/sessions/contrib/<sessionid>(/path?) and the backend should listen for requests at /<path>, correct?

@BearBearCodes Yes, that's right, I did restore the middleware so that the server doesn't need to care about the internal path to listen on.

brianmajor:
> First, I think as much as possible of the launched configuration should be exposed through standardized environment variables. We have some now for e.g. the number of cores that are specific to different software. These could become CADC_NUM_CORES for example.

Okay, good idea.  Currently the following skaha-specific environment vars are set at runtime.  There was:

```
        env:
        - name: skaha_hostname
          value: "${skaha.hostname}"
        - name: skaha_username
          value: "${skaha.userid}"
        - name: skaha_sessionid
          value: "${skaha.sessionid}"
```
And I just added these (HOME should have been there to begin with):
``` 
        - name: HOME
          value: "/arc/home/${skaha.userid}"
        - name: PWD
          value: "/arc/home/${skaha.userid}"
        - name: JULIA_NUM_THREADS
          value: "${software.requests.cores}"
        - name: OPENBLAS_NUM_THREADS
          value: "${software.requests.cores}"
        - name: MKL_NUM_THREADS
          value: "${software.requests.cores}"
        - name: OMP_NUM_THREADS
          value: "${software.requests.cores}"
```
Anything else?

> Second, could we use this opportunity to standardize the launch process a bit? For example respecting the image CMD and ENTRYPOINT instead of using startup.sh. These could of course point to a startup script if desired.
Similarly I’d be interested if there’s a way to avoid copying those setup files into the image.

Another good idea, as long as the ENV from the launch overrides ENV from the Dockerfile.  I think it does?  That's important so that people can test without the platform using their own ENV.

> With those changes it would become much easier to adapt any docker image exposing a web service.

👍

BearBearCodes:
Hi, I'm just wondering if there is an ETA for when we can expect these changes to be merged on ws-CANFAR

brianmajor:
Hi Isaac.  Sorry for the delay.  I'm hoping to put this on our test host
for you by the end of the week.  How do the changes look so far?  Any other
updates?

BearBearCodes:
No worries @brianmajor. The only change I had to make was to modify one line of my Flask code to point to "`/`" instead of "`session/castor-etc/<session-id>`" since we are bringing back the "`middleware`". Once you've put it on the test host, just let me know and I'll make a new image and test it ASAP.

brianmajor:
Isaac, I forgot to make the change to the startup procedure as suggested by
William, but it's there now.  It means that you should ensure that the
ENTRYPOINT in your runs your /skaha/startup.sh script.  startup.sh can get
the session id (if needed) from the `skaha_sessionid` environment variable.

BearBearCodes:
Okay thank you for letting me know. I think I have implemented the `ENTRYPOINT` correctly to accept the Docker `skaha_sessionid` environment variable (following https://stackoverflow.com/a/67206046):

```dockerfile
ENV skaha_sessionid=1
ENTRYPOINT [ "/bin/bash", "-c", "exec /skaha/startup.sh ${skaha_sessionid} \"${@}\"", "--" ]
```

But it is not critical to get the session ID (I'm doing this just to test if this approach works). Since the `middleware` component is back, I can always default back to just:

```dockerfile
ENTRYPOINT [ "/skaha/startup.sh" ]
```

Once these changes are up on rc-CANFAR, I will test the image to make sure the CASTOR ETC GUI works. I'll keep you updated.

brianmajor:
> Okay thank you for letting me know. I think I have implemented the `ENTRYPOINT` correctly to accept the Docker `skaha_sessionid` environment variable (following https://stackoverflow.com/a/67206046):
> 
> ```dockerfile
> ENV skaha_sessionid=1
> ENTRYPOINT [ "/bin/bash", "-c", "exec /skaha/startup.sh ${skaha_sessionid} \"${@}\"", "--" ]
> ```
Hmm, I'm not completely certain, but I don't think that will work.  I don't think docker/k8s will dynamically replace entrypoint arguments based on environment vars.  Instead I would, from within startup.sh, get the sessionid from the ENV var like so:
```
$skaha_sessionid
```
That var should exist already, so it's actually quite simple and no need for params.

> 
> But it is not critical to get the session ID (I'm doing this just to test if this approach works). Since the `middleware` component is back, I can always default back to just:
> 
> ```dockerfile
> ENTRYPOINT [ "/skaha/startup.sh" ]
> ```
Yes, this is probably sufficient.

> Once these changes are up on rc-CANFAR, I will test the image to make sure the CASTOR ETC GUI works. I'll keep you updated.

You can test the ENV stuff with `docker run` by using the -E param to make sure session_id is overrides the one from your Dockerfile.

BearBearCodes:
Oh okay I understand. I'll try your approach of getting the `$skaha_sessionid` environment variable within `startup.sh`. Regardless, it is not a big deal if this doesn't work. I'll let you know how the testing goes sometime tomorrow since I have a class that starts in 1 hr and I need to eat first since it runs until 10 pm... :')

(and I have a class that starts at 8:30 am tomorrow! Yay!)

brianmajor:
Hi @BearBearCodes - hope your classes went well.  I have these changes on https://rc-uv.canfar.net so you can test (when you find some time :)  There's a new label in https://images.canfar.net called 'contributed'.  You'll need to label your updated image with that label.  Let me know if you find any issues or if need a reminder of how to launch images on rc-uv.  Cheers.

BearBearCodes:
Hi @brianmajor, thanks for your help on this. I have tested the GUI on my local machine and it works, but when I try to launch a session on rc-uv.canfar.net, I get the following error:

```bash
unexpected exception: java.nio.file.NoSuchFileException: /root/config/launch-contributed-etc.yaml
```

brianmajor:
@BearBearCodes  Ooops, a bug, thanks.  Fixed and deployed again on rc-uv.

BearBearCodes:
I now get a different error:
```bash
unexpected exception: java.io.IOException: Error executing command: [kubectl, create, --namespace, skaha-workload, -f, /tmp/9efb34a3-8400-4000-873e-2bedadc5449c] Error: error: error parsing /tmp/9efb34a3-8400-4000-873e-2bedadc5449c: error converting YAML to JSON: yaml: line 83: did not find expected key
```

brianmajor:
Thanks again and fixed again.  I was able to launch an instance of your image but I didn't test or try the session URL that was created.

BearBearCodes:
HI @brianmajor, my session on rc-uv.canfar.net has been stuck in "pending" for about 10 minutes now.

The last line of the Dockerfile is

```dockerfile
ENTRYPOINT [ "/skaha/startup.sh" ]
```

and the `/skaha/startup.sh` script is executable (using `chmod a+x`).

The `startup.sh` script itself is just:

```bash
#!/bin/bash

skaha_sessionid=${1:-test123}

echo "Starting gunicorn..."

cd /backend
gunicorn -b 0.0.0.0:5000 connector:app \
        --log-level=debug \
        --log-file=/dev/stdout \
        -e skaha_sessionid=${skaha_sessionid}
```

The container launched using the Docker image works perfectly on my local machine. But on rc-CANFAR, I can't even see the session logs to debug what is going on because it is stuck on "Pending". Do you have any idea what could be going on (maybe the `ENTRYPOINT` is not being respected)?

EDIT: for the record, I don't see any "Starting gunicorn..." message either.

brianmajor:
I found another bug on our side.  Fixed now and the containers are reaching the 'running' state.

BearBearCodes:
Hi @brianmajor. Thanks for the quick fixes! The ETC GUI looks to be working now 😀! When is the expected deployment of the "contributed" session type on ws-uv.canfar.net?

BearBearCodes:
Hi @brianmajor, I have a quick technical question that just came to mind. Since you restored the `middleware` component in the k8s ingress, would that mean multiple users' requests would be directed to the same location (i.e., root `/`), thereby mixing up multiple users' requests?

Suppose I had the CASTOR web app running in a container on my account and you also had another instance of the web app running in another container on your account: if you submit a request, would my container also pick up on your request?

To test this, I have an ETC container running on rc-uv.canfar.net. You can try launching a `castor_etc_gui:1.0.0` container and submitting a Telescope request via the green "SAVE" button at the bottom of the page (you don't need to modify any of the default form inputs). Once you do that, let me know and I will check my session logs to see if Flask thinks I made a request to the server.

brianmajor:
Hi @BearBearCodes - Fantastic.  As for your question about multiple sessions:  I hit 'save' in a new session I created, but I'm certain we don't have to worry about this.  Each 'launch' creates a new instance of the images with it's own session id.  The middleware simply does the routing from URL (that contains the session id) to the container associated with that session id.  Unless there is something that castor-etc does at the shared file system level (ie in /arc/) that is not specific to the session id, they are completely isolated.  Anyhow, let me know if it still looks good on your side and I will release.

As for the portal, I hope that type 'contributed' will be on there shortly after, hopefully this week.

BearBearCodes:
Hi @brianmajor, thank you so much for all your help! I can confirm that my session logs still look the same (i.e., no `Telescope` requests were made) 😃. Thanks again!

P.S. my replies may be a bit delayed today since I have class for the next 6 straight hours.

brianmajor:
Okay, great, I'll release this now.

brianmajor:
Done.  Looks okay to me on https://ws-uv.canfar.net/skaha

BearBearCodes:
@brianmajor, follow-up question: should we be able to see the "contributed" session type under the drop-down menu? I can only see "notebook", "carta", "desktop", and "pluto".

brianmajor:
Hi Isaac.  No, the type needs to be added to the portal.  That's the part I hope will be done by the end of the week.

BearBearCodes:
Ah okay. Thanks for the clarification!

brianmajor:
Hi @BearBearCodes - The portal changes have been released, so there's now a type 'contributed'  in the drop-down list.  (I had to do a hard-refresh on my browser for some reason to get it to show up.)  It looks like the label on the castor image is no longer there now though, so no images appear for that type.  Could you please put the label back when you're ready?

Also, I wonder if you want to call your image forecastor instead of castor-etc-gui?  It might be more familiar now to users?

BearBearCodes:
HI @brianmajor, thank you for the release! I can confirm that the `castor_etc_gui` image can be launched via the dropdown menu and there doesn't seem to be any major issues. I've thought about naming it `forecastor_etc` or `forecastor_etc_gui`, but our Python package (and the notebook image on images.canfar.net) is named `castor_etc`, so I figured we should keep the naming consistent for the GUI portion, hence `castor_etc_gui`.

I will be notifying those involved with CASTOR about this release and make all our code public now. Thanks again @brianmajor!

EDIT: released!

brianmajor:
Okay, great to hear @BearBearCodes .  It's been great working with you on this.  The exercise has resulted in improvements to the platform for everyone.

BearBearCodes:
Thank _you_ for all your help throughout this project, @brianmajor! Your advice from the early stages of the project really made integrating the app as smooth as possible. Not to mention your new `contributed` type simplified the GUI backend (e.g., now it is agnostic to the session ID and connection path). Also thanks @sefffal for your suggestions regarding the new session type.

sefffal:
Hi @brianmajor! Thanks for implementing this feature. I'm just getting a chance to test it out now.

I'd like to start by migrating my pluto image to a contributed session type. I've altered `ENTRYPOINT` as follows:
```docker
ENTRYPOINT [ "/skaha/startup.sh"]
```

A couple questions. First, is `CMD` supported as well? Second, I have 
```docker
EXPOSE 1234
``` 
to expose the Pluto service on it's usual port. Do I need to alter this to match @BearBearCodes's  port 5000 or does the platform know what port I expose?

Thanks for your help!

brianmajor:
Hi @sefffal !  Great, glad to hear you are taking this on.  It should be pretty straight forward.

The 'contributed' session type now completely defers startup to the container.  So you can do by defining that ENTRYPOINT (/skaha/startup.sh) or any other way.  CMD would also work.

The port however must be 5000.  That's a neat idea to have the platform pick it up but not sure it's possible to do that level of introspection.

You'll have to change the label in harbor from 'pluto' to 'contributed' to test.

Let me know if you have any more questions.

sefffal:
Thanks, that's working perfectly right out of the box. Shall I create a PR to update the container definition in the skaha repo?

PR #334, opened 2022-05-25
Change session type of default pluto container to `contributed` from special case `pluto`

sefffal:
Change session type of default pluto container to `contributed` from special case `pluto`.
See: https://github.com/opencadc/skaha/pull/323


brianmajor (approved):

brianmajor:
Thanks @sefffal , glad it wasn't much work.  We will remove the type 'pluto' from the option list on the portal.  Thanks again for your contributions.

PR #344, opened 2022-06-08
Cadc 11448 Removed skaha/desktop dependency on skaha/terminal version

yeunga:


deployment/k8s-config/kustomize/base/skaha-workload/config/build-menu.sh
```diff
@@ -2,29 +2,52 @@
 
 HOST=$1
 STARTUP_DIR="/desktopstartup"
-EXECUTABLE_DIR="$HOME/.local/bin"
-DESKTOP_DIR="$HOME/.local/share/applications"
-DIRECTORIES_DIR="$HOME/.local/share/desktop-directories"
+OLD_EXECUTABLE_DIR="$HOME/.local/bin"
+OLD_DESKTOP_DIR="$HOME/.local/share/applications"
+OLD_DIRECTORIES_DIR="$HOME/.local/share/desktop-directories"
```
brianmajor: Should these 'OLD' vars be removed now?

deployment/k8s-config/kustomize/base/skaha-workload/config/build-menu.sh
```diff
@@ -2,29 +2,52 @@
...
+    if [[ -d "$dir" ]]; then
+      # delete the directory
+      rm -rf $dir
+    fi
+  done
+}
+
+init () {
+  if [[ ! -d "${SKAHA_DIR}" ]]; then
+    delete_old_dirs
```
brianmajor: I wouldn't worry about automatic cleanup.  I means we would have to have a second release with this code removed in the future.  I think it's harmless to leave the old dirs there.

brianmajor (changes requested):

deployment/k8s-config/kustomize/base/skaha-workload/config/build-menu.sh
yeunga: No, these 'OLD' vars were added to change the directory structure from $HOME/.local/ to $HOME/.local/skaha. They need to be there until all current users have executed the script once, i.e. have changed their directory structure.

deployment/k8s-config/kustomize/base/skaha-workload/config/build-menu.sh
yeunga: It's trivial to remove the old dirs now. If we don't remove them now and if some users add their stuff to these dirs, it would be more difficult to clean up the dirs. These old dirs are what the current users are using.

yeunga: Yes, we'll have a second release that removes the codes that handle the old dirs.

brianmajor: Okay, but when would this second release occur?  Would we have to check user's home directories and see if it's been done?  What about simply removing those old scripts now manually?  I might be a good time since all sessions were stopped yesterday because of the upgrade.

yeunga: Sure, we can remove those dirs now. We need to ensure that we release this fix before the users start up a new session after the upgrade.

brianmajor: Okay, perfect, let's move towards release then.  I guess you can remove the cleanup from the scripts.

yeunga: So I'll remove the code to delete the old dirs from build-menu.sh. We'll need to run a separate script to remove the old dirs. We'll release this fix before the users start up a new session.

yeunga: Do we have time for the above?

brianmajor: Sure, lots of time. If we have to skip a few users that's okay.

deployment/k8s-config/kustomize/base/skaha-workload/config/build-menu.sh
```diff
@@ -107,15 +125,22 @@ build_menu_item () {
   sed -i -e "s#(IMAGE_ID)#${image_id}#g" $executable
   sed -i -e "s#(NAME)#${name}#g" $executable
   sed -i -e "s#(NAME)#${name}#g" $desktop
-  sed -i -e "s#(HOME)#$HOME#g" $desktop
+  sed -i -e "s#(EXECUTABLE)#${EXECUTABLE_DIR}#g" $desktop
   sed -i -e "s#(CATEGORY)#${category}#g" $desktop
+  if [[ ${name} == *"terminal:"* ]] && [[ "${name}" > "${TERMINAL_VERSION}" ]]; then
```
brianmajor: I think you'd better match on more that just 'terminal' because there could be user images that contain that name.  We control the skaha project in harbor, so match on `*/skaha/terminal:*`?

brianmajor (changes requested):

deployment/k8s-config/kustomize/base/skaha-workload/config/build-menu.sh
yeunga: Changed.

brianmajor (approved):

PR #352, opened 2022-06-16
CADC-11448 Updated build-menu.sh to use relative soft link.

yeunga:


brianmajor:
We've agreed this isn't a valuable change.

PR #363, opened 2022-06-23
CADC-11550 Updated algorithm that generates the job name to allow for…

yeunga:
… k8s label length restriction.

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
+++ b/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
@@ -470,7 +470,7 @@ public class PostAction extends SessionAction {
         String param = name;
         log.debug("Using parameter: " + param);
         
-        String uniqueID = new RandomStringGenerator(8).getID();
+        String uniqueID = new RandomStringGenerator(3).getID();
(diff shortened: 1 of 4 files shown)


skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -467,9 +471,20 @@ public void attachSoftware(String image, List<String> params) throws Exception {
         log.debug("Using parameter: " + param);
         
         String uniqueID = new RandomStringGenerator(8).getID();
```
brianmajor: Could you please change this uniqueID to be length 3 rather than 8.  It will save us some space.

brianmajor (changes requested):

brianmajor:
When you look at launchSession, you will have to be sure that the associated ingresses and services are using the same variable to identify the job name so they continue to connect to each other.  DeleteAction should be checked to make sure it's deleting the right thing as well.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
yeunga: Done

brianmajor (approved):

PR #366, opened 2022-06-30
CCADC-11578 Updated image/GetAction to improve performance.

yeunga:


skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
```diff
@@ -315,7 +315,6 @@ protected String callHarbor(String idToken, String harborHost, String project, S
         } else {
             harborURL = new URL("https://" + harborHost + "/api/v2.0/projects/" + project + "/repositories/"
                 + repo + "/artifacts?detail=true&with_label=true&page_size=-1");
-            message = "artifacts";
```
brianmajor: Why was this removed?

brianmajor (changes requested):
Looks good Alinga, just the one question/comment.

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
yeunga: Good eye. It was accidentally removed. Added back.

brianmajor (approved):

PR #369, opened 2022-07-22
Cadc 11492 Added affinity specs to reduce GPU usage

yeunga:


deployment/k8s-config/harbor/values-keel-dev.yaml
```diff
@@ -78,7 +143,13 @@ persistence:
 #***************************************************
 core:
   priorityClassName: uber-user-preempt-high
```
brianmajor, line 152: I think this priority class name needs to be added to each component of harbor, not just core.

deployment/k8s-config/kustomize/base/arc/arc-sshd-deployment.yaml
```diff
@@ -15,6 +15,13 @@ spec:
       labels:
         run: arc-sshd
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: nvidia.com/gpu.count
+                operator: DoesNotExist
```
brianmajor, line 24: priority class name should be added to this spec (and to others that don't have it already)

deployment/k8s-config/kustomize/base/arc/cavern-tomcat-deployment.yaml
```diff
@@ -43,6 +43,7 @@ spec:
           readOnly: true
       securityContext:
         runAsUser: 0
+      priorityClassName: uber-user-preempt-high
       serviceAccountName: skaha
       volumes:
       - name: servops-cert
```
brianmajor: this file is missing the gpu affinity

brianmajor (changes requested):
In general the affinity and priority class names need to be contained in all of these files if not there already.

brianmajor (changes requested):
Didn't you need to update the harbor chart version?  That should be part of this PR because it needs to be released on prod.

deployment/k8s-config/harbor/values-keel-dev.yaml
yeunga: Updated each harbor component with both affinity and priority class. The priority class of cadc-harbor-jobservice is not set even though it is in the spec.

deployment/k8s-config/kustomize/base/arc/arc-sshd-deployment.yaml
yeunga: Added and verified.

deployment/k8s-config/kustomize/base/arc/cavern-tomcat-deployment.yaml
yeunga: Removed all cavern related files and directories as cavern has been superseded by arc.

brianmajor (approved):

PR #371, opened 2022-08-09
Cadc 11684 Added error exit codes and handling of error status.

yeunga:
Also removed extra codes in vnc_startup.sh

### DIFF
--- a/containers/session-containers/skaha-desktop/apply-version.sh
+++ b/containers/session-containers/skaha-desktop/apply-version.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 . VERSION && echo "tags: $TAGS"
 for t in $TAGS; do
-   docker image tag images.canfar.net/skaha/desktop:latest images.canfar.net/skaha/desktop:$t
+   docker image tag images-rc.canfar.net/skaha/desktop:latest images-rc.canfar.net/skaha/desktop:$t
 done
 unset TAGS
-docker image list images.canfar.net/skaha/desktop
+docker image list images-rc.canfar.net/skaha/desktop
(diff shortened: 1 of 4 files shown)


containers/session-containers/skaha-desktop/apply-version.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 . VERSION && echo "tags: $TAGS"
 for t in $TAGS; do
-   docker image tag images.canfar.net/skaha/desktop:latest images.canfar.net/skaha/desktop:$t
```
brianmajor: This should probably remain as images, not images-rc.  Or you could make it do both.

containers/session-containers/skaha-desktop/src/common/scripts/vnc_startup.sh
```diff
@@ -37,9 +37,6 @@ Fore more information see: https://github.com/ConSol/docker-headless-vnc-contain
 if [[ $1 =~ -h|--help ]]; then
     help
     exit 0
-"vnc_startup.sh" 122L, 4274C
-    help
-    exit 0
```
brianmajor, line 42: It would be better if no change to the container was made.  Is this necessary?  What does it accomplish?

brianmajor (changes requested):
Were you able to test this?  You could hard code it to fail (temporarily) to test maybe?

yeunga:
> Were you able to test this? You could hard code it to fail (temporarily) to test maybe?

Yes, I tested the changes.

containers/session-containers/skaha-desktop/apply-version.sh
yeunga: Right. Changed back to images.

containers/session-containers/skaha-desktop/src/common/scripts/vnc_startup.sh
yeunga: This was the section of code that I mentioned to you in slack that I did not understand what it was doing and thought it was extra code. Just want to clean up the code. Not related to my other changes.

brianmajor (approved):

PR #373, opened 2022-08-25
Cadc 11204

kcasteels:
There are helper scripts for getting cluster stats and for setting quotas.

### DIFF
--- a/deployment/k8s-config/scripts/getquotas.sh
+++ b/deployment/k8s-config/scripts/getquotas.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+PATH=/cephfs/cavern/home
+re='^[0-9]+$'
+
+for directory in $PATH/*; do
+    echo $directory
+    QUOTA=$(/usr/bin/getfattr --only-values $directory)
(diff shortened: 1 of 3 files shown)


brianmajor (commented):
Before getting into the details, I'm wondering how getquotas.sh and setquotas.sh can be executed in their current form.  They need to have access to cephfs and be run as root.  Connecting to a system container to do this and then copying in the script really takes away the automation.  Perhaps it should follow a pattern like add-user and add-project, where there are small containers that run and do the work.  See:

https://github.com/opencadc/science-platform/blob/master/deployment/ops-containers/add-user/Dockerfile

kcasteels:
The original quotas task required changing the directory quotas manually as a once off thing. I wrote these scripts to help me in that task. If we plan to change quotas regularly then it makes sense to have them in a container like for add-user, but if its a rarely used once off thing, then having the scripts in their current form is useful. It allows us to view the output as directories are analyzed and react accordingly.

brianmajor:
Okay, I'm fine with that.  Thanks for the PR.

brianmajor (approved):

PR #378, opened 2022-09-08
CADC-11749 Removed code to create the data symlink. This is now done …

yeunga:
…in the Dockerfile.

### DIFF
--- a/deployment/k8s-config/kustomize/base/skaha-workload/config/start-desktop-software.sh
+++ b/deployment/k8s-config/kustomize/base/skaha-workload/config/start-desktop-software.sh
@@ -3,14 +3,6 @@
 USERID=$1
 TITLE=$2
 
-if [[ -z "${CASA_RELEASE}" ]];
-then
-    echo "not a casa container"
-else
-    echo "creating casa data repo link"
-    ln -s /arc/projects/casa-data-repository/ /opt/${CASA_RELEASE}/data
-fi
-
 echo "[skaha] Starting skaha software container [title=$TITLE] for [userid=$USERID]"
 
 INITFILE=/skaha/init.sh


brianmajor (approved):

brianmajor:
@helenkirk :  @yeunga has moved creation of the symbolic link to the CASA leap second data from the skaha startup script (the change in this PR) to be in the containers themselves.  (In the casa Dockerfiles.).   This is a better design because it removes a coupling between CASA and the platform.  However, once we release this change, the current CASA containers will no longer work (won't have the leap second data) because they have not been rebuilt and published with the Dockerfile changes.  So, we will hold off merging and releasing this PR until that has been done.

helenkirk:
Thanks for the heads-up, @brianmajor .  For clarification, @yeunga, will it be sufficient to port those commands that had been in the skaha startup script into the casa container docker file, or am I going to need to put in some additional steps as well?

brianmajor:
@helenkirk I think they just need to be rebuilt and published without any more changes.  @yeunga can confirm.

helenkirk:
I can confirm the changes work with a rebuilt casa6.4.0 container.

yeunga:
@helenkirk Thank you for confirming that a rebuilt casa6.4.0 container works. There is no need to port those commands that had been in the skaha startup script into the casa container docker file. All needed changes to the docker file have been done. The containers just need to be rebuilt and published. 

w.r.t. my Makefile changes, I did not update the Makefile in version-3.4-4.2. The Makefile needs to be updated for the new AstroSoftware->casa-[3|4] menu.

helenkirk:
Thanks @yeunga I'll make sure that I update the Makefile for those casa versions before re-compiling those casa containers.  I appreciate you having make all of the necessary updates to the Docker files already!

brianmajor:
Hey @helenkirk - I think I remember that you finished updating all the casa containers but thought I'd check before merging this change in.

helenkirk:
@brianmajor They are mostly done, and I think it would be fine to integrate it at this stage.  I've now incorporated the changes & tested new containers from 4.3.0 through to 6.4.3.  I still have to tackle the final casa container directory (3.4 to 4.2), but thus far no problems with the database change have come up, so it presumably should be fine.  I'm hoping to get started on this final directory soon.

brianmajor:
Yes, okay, I think we've proven that the approach works.  I'll merge it.  Thanks @helenkirk

PR #385, opened 2022-10-07
change session limits to be by count, not type

brianmajor:


skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
```diff
@@ -136,12 +137,20 @@ public SkahaAction() {
         }
         skahaUsersGroup = System.getenv("skaha.usersgroup");
         skahaAdminsGroup = System.getenv("skaha.adminsgroup");
+        String maxUsersSessionsString = System.getenv("skaha.maxusersessions");
```
yeunga, line 140: Do we need to add skaha.maxusersessions to skaha-tomcat-deployment.yaml?

yeunga (changes requested):
Looks good other than the comment maxusersessions.

yeunga (approved):

yeunga (approved):

yeunga (approved):

PR #401, opened 2022-11-10
Substantially updated documentation

helenkirk:
This pull request contains all of the updated documentation for new users, including new tutorials on methods of transferring files in & out of the science portal, and a new overview page that should reduce our need for typing instructions when onboarding new users.


helenkirk:
A side note for Brian: these new documentation pages include updating all of the old info at https://github.com/canfar/arcade/blob/master/arcade-tutorial/README.md that we wanted to phase out.  The one exception to note is the detailed instructions under "How to set up CADC's vos tools to transfer from your local computer to arcade" on that link which I did not include as I think it's no longer applicable (e.g., having to manually create and edit a VOS config file before getting started).

brianmajor:
Hi @helenkirk - Yes, you are right about the VOS config file.  It no longer needs to be edited for use with /arc.

Do you plan on having someone review this (try it out?) or should I merge it in?

Oh, and thank you for this work.

helenkirk:
Hi Brian,

I'm not sure how to really open this up for a good review without it being published.  On github (which anyone could look at from my page), individual webpages seem to be fairly readable, but the main landing page, links between pages, etc, don't seem viewable.  I built & tested everything locally, but that's only viewable as .html files on my computer, not online.  If you're okay with just merging it, Brenda mentioned that one of her collaborators would be keen to use the material in the file transfer series once it is live, and that is where the majority of updates have happened, so that could be an after the fact test.  And our next user to be onboarded would be directed to the other major update (overview of the system).  It does raise an interesting point about testing for the documentation in general... happy to discuss this further with you if you have any thoughts on an optimal approach...

brianmajor:
Yes, no problem.  I will merge this and we can deal with any future changes as new PRs.

brianmajor:
Okay, all done.

helenkirk:
Thanks!

PR #403, opened 2022-11-22
Cadc 11901a Rework

yeunga:
- based on Brian's review comments
- refactored SessionAction to improve performance
- added session ID to desktop-app
- fixed DeleteAction which exposed a bug as a result of adding session ID to desktop-app


deployment/k8s-config/kustomize/base/skaha/config/launch-desktop-app.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   template:
     metadata:
       labels:
+        canfar-net-sessionID: "${skaha.sessionid}"
```
brianmajor, line 16: For symmetry, this label should also be added to the label section above.  (and in the overlay version of this file)

skaha/src/main/java/org/opencadc/skaha/session/DeleteAction.java
```diff
@@ -105,24 +105,33 @@ public void doAction() throws Exception {
...
+                        // want to ignore them as we pick the session to be deleted.
+                        for (String line : lines) {
+                            String[] parts = line.split("\\s+");
+                            if (!TYPE_DESKTOP_APP.equals(parts[0])) {
+                                String sessionUserId = parts[1];
+                                if (!userID.equals(sessionUserId)) {
+                                    throw new AccessControlException("forbidden");
+                                }   
+    
+                                String type = parts[0];
```
brianmajor: The type could be extracted (this line) before it's used (parts[0]) on line 116.

brianmajor (changes requested):
Looking good, just a few minor comments

deployment/k8s-config/kustomize/base/skaha/config/launch-desktop-app.yaml
yeunga: Good eye. Added.

skaha/src/main/java/org/opencadc/skaha/session/DeleteAction.java
yeunga: Updated.

brianmajor (approved):

PR #404, opened 2022-11-23
bug fix: don't include desktop-apps in session count

brianmajor:


yeunga (dismissed):
Looks good to me. Need to update the skaha-tomcat-deployment files I suppose.

brianmajor:
Ah, right, thanks for the reminder.

PR #405, opened 2022-11-23
Cadc 11734a Refactored session/GetAction to improve performance.

yeunga:


skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
```diff
@@ -176,86 +179,125 @@ private ResourceStats getResourceStats() throws Exception {
...
 
-            String withRAMStr = String.valueOf(withRAM) + "Gi";
-            String maxRAMStr = String.valueOf(maxRAM) + "Gi";
+            // convert RAM unit from Ki to Gi
+            String withRAMStr = String.valueOf(withRAM/1048576) + "Gi";
+            String maxRAMStr = String.valueOf(maxRAM/1048576) + "Gi";
             return new ResourceStats(desktopCount, headlessCount, totalCount, coresInUse, coresAvailable, maxCores, withRAMStr, maxRAMStr, withCores);
         } catch (Exception e) {
             log.error(e);
             throw new IllegalStateException("failed reading k8s-resources.properties", e);
```
brianmajor: I think this is the wrong error message.  Something like "failed gathering statistics" would be better.

skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
```diff
@@ -176,86 +179,125 @@ private ResourceStats getResourceStats() throws Exception {
...
                 }
 
-                if (!hasRAM && line.indexOf("memory:") >= 0) {
-                    String[] parts = line.split(":");
-                    // amount of RAM from "Capabity.memory" in Ki
-                    int ram = Integer.parseInt(parts[1].replaceAll("[^0-9]", "").trim())/1048576;
-                    log.debug("Available memory in node " + nodeName + ": " + ram + " Gi");
-                    resources[1] = ram;
-                    hasRAM = true;
+                if (hasName || hasCapacity || hasCores) {
```
brianmajor: Perhaps the logic is right, but it doesn't read very well.  You'd think this if statement means that you have at least one of name, capacity, or cores.  But then an exception is thrown saying you don't have those, so is confusing.

skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
```diff
@@ -176,86 +179,125 @@ private ResourceStats getResourceStats() throws Exception {
...
+        String getAvailableResourcesCmd = "kubectl -n " + k8sNamespace + " describe nodes ";
+        String rawResources = execute(getAvailableResourcesCmd.split(" "));
+        Map<String, int[]> nodeToResourcesMap = new HashMap<String, int[]>();
+        if (StringUtil.hasLength(rawResources)) {
+            String[] lines = rawResources.split("\n");
+            if (lines.length > 0) {
+                List<String> keywords = Arrays.asList("Name:", "Capacity:", "cpu:", "memory:");
+                String nodeName = "";
+                boolean hasName = false;
+                boolean hasCapacity = false;
```
brianmajor, line 246: What is 'capacity'?  memory?

brianmajor (commented):
Just a few comments for now... I'm going to ask you to walk me through some of the logic. :)

skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
yeunga: Good eye. Updated exception message.

skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
yeunga: I am going to add a comment to explain this check. Basically at this point, we have processed all nodes. All flags should have been cleared. If not, then there is a problem. Either the output from kubectl is incorrect (e.g. missing some lines) or there is a problem in our logic.

skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
yeunga: They are in the output of "kw describe nodes". As we process each line in the output, we want to know where are in the output. We want 'Capabity.cpu' and 'Capabity.memory'; however we do not want 'Allocatable.cpu' and 'Allocatable.memory'.

skaha/src/main/java/org/opencadc/skaha/session/GetAction.java
yeunga: Code removed after the latest update.

brianmajor (approved):

PR #406, opened 2022-11-24
Cadc 11901b Updated handling of 'metrics not available' IOException

yeunga:


skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -453,12 +453,7 @@ private Map<String, String[]> getResourceUsages(String k8sNamespace, String forU
                 }
             }
         } catch (IOException ex) {
-            if (ex.getMessage().contains("Metrics not available")) {
-                // no session using any resources, return empty resourceUsages
-            } else {
-                // other errors, propagate the exception
-                throw ex;
-            }
+            // error or no session using any resources, return empty resourceUsages
```
brianmajor, line 456: Could you please put a log.debug() message in here so we can diagnose errors if needed?
```log.debug("failed to query for metrics", ex);```

brianmajor (changes requested):

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: Added

brianmajor (approved):

PR #415, opened 2022-12-06
Cadc 11880a Added keep-alive support to Skaha API

yeunga:
Also updated the swagger doc. However the swagger doc has not been tested. Not sure how to display the swagger doc on keel-dev.


skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -214,7 +216,19 @@ public void doAction() throws Exception {
...
+                    List<String> jobNames = getRenewJobNames(userID, sessionID);
+                    if (!jobNames.isEmpty()) {
+                        for (String jobName : jobNames) {
+                            renew(jobName);
+                        }
+                    } else {
+                        throw new IllegalArgumentException("No active job for user " + userID + " with session " + sessionID);
+                    }
+                } else {
+                    throw new UnsupportedOperationException("Cannot modify an existing session.");
```
brianmajor: I think this last `else` should be one level out (ie after line 232), not here.  This should stay however but say "unrecognized action" instead.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -238,6 +252,69 @@ public void doAction() throws Exception {
...
+            renewExpiryTimeCmd.add("patch");
+            renewExpiryTimeCmd.add("job");
+            renewExpiryTimeCmd.add(jobName);
+            renewExpiryTimeCmd.add("--type=json");
+            renewExpiryTimeCmd.add("-p");
+            renewExpiryTimeCmd.add("[{\"op\":\"add\",\"path\":\"/spec/activeDeadlineSeconds\", \"value\":" + activeDeadlineSeconds + "}]");
+            
+            execute(renewExpiryTimeCmd.toArray(new String[0]));
+        } else {
+            throw new IllegalStateException("failed to obtain session expiry time");
```
brianmajor: Message here should be: `"missing configuration item: " + SKAHA_SESSIONEXPIRY`

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -374,10 +374,15 @@ public void streamPodLogs(String forUserID, String sessionID, OutputStream out)
     public Session getSession(String forUserID, String sessionID) throws Exception {
```
brianmajor, line 374: Is getSession used elsewhere?  I just want to confirm that this won't change the behaviour of the skaha API call to /skaha/session (ie list sessions).

brianmajor (changes requested):

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: SessionAction.getSession(...) is used by GetAction.getSingleSession(...) only so far.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
yeunga: Updated.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
yeunga: Updated message to "unrecognized action".

Added a check for the existence of "action" and move the above-mentioned UnsuppoertedOperationExcetion to the else clause of this check.

brianmajor:
Hi @yeunga - I was playing around with the new "action=renew" functionality on keel-dev but have found a problem:  The expiry time of the job is not changing after the patch is applied.  My guess is that the 'count' of the `activeDeadlineSeconds` is not restarted after the patch is applied.  That probably means that you have to do a little math for the correct value.  Something like:
```newActiveDeadlineSeconds = (now - jobStartTime) + initialActiveDeadlineSeconds```
@kcasteels -- could you please inform @yeunga how this is done with your scripts?
There's a chance that my testing was incorrect as well, but I'm pretty sure I did not see the `expiryTime` (a timestamp field) change after doing the renew action.

**Comment by yeunga** [comment]
Hi @brianmajor When I tested it during development, I could see the activeDeadlineSeconds was changed to the set value. I'll test it again.

**Comment by yeunga** [comment]
Looks like the problem is that I implemented getSessionExpiryTimeCMD(...) to get the spec.activeDeadlineSeconds as the current expiry time. However this value does not seem to count down or change. We need a copy of this configured value which gets updated; the same way that Kubernetes keeps track of the expiry time and ejects the job when its time expires.

**Comment by yeunga** [comment]
Updated expiry time renewal algorithm. We now extend the expiry time by a multiple of the configured activeDeadlineSeconds. The expiry time will be less than 2 * activeDeadlineSeconds.

File: skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java [hunk 4]
```diff
@@ -416,6 +421,8 @@ private List<Session> getSessions(String forUserID, String sessionID) throws Exc
                         } else {
                             instant = instant.plus(Integer.parseInt(expiryTimesStr), ChronoUnit.SECONDS);
                         }
+                        
+                        session.setExpiryTime(instant.toString());
```
brianmajor: Looks like this will override the setting on <none> on line 420?

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -374,10 +374,15 @@ public void streamPodLogs(String forUserID, String sessionID, OutputStream out)
...
         List<Session> sessions = getSessions(forUserID, sessionID);
         if (sessions.size() >0) {
-            return sessions.get(0);
-        } else {
-            throw new ResourceNotFoundException("session " + sessionID + " not found");
-        }
+            for (Session session : sessions) {
+                // exclude 'desktop-app'
+                if (!SkahaAction.TYPE_DESKTOP_APP.equalsIgnoreCase(session.getType())) {
+                    return session;
```
brianmajor, line 380: This isn't related to action=renew, so does it fix another separate issue?

brianmajor (changes requested):

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: Yes. When we do not find an expiry time for the uid, we set the expiry time to "none", otherwise we set the expiry time to the time that we calculated.

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: Yes, there was a bug that I introduced that given a desktop session ID, skaha returned the desktop session as well as the desktop apps. The code only picked the first item returned which turned out to be almost always a desktop-app. This fixed the bug. I mentioned the bug fix and you were okay with putting the fix in this release.

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
brianmajor: But this line is called no matter what, so the "none" logic never sticks.

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
brianmajor: Okay, yes, thanks for the reminder.  Do you know if this will have implications on the portal?

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: Oh... right. Good eyes. I'll fix it. Thanks.

brianmajor (approved):

PR #420, opened 2022-12-19
Cadc 12067 Use number of cores instead of millicores ('m') and use 'T', 'G', 'M', 'K' instead of 'Ti', 'Gi', 'Mi', 'Ki'

yeunga:


brianmajor (changes requested):
@yeunga I would like to see some unit tests for these conversion functions.  It would be easier to understand what they're doing, and since they're so isolated, they make good candidates for unit tests.  You can make the method signatures non-private where needed to make them visible to the test code.  Only write them for ones that don't require kubernetes interaction.

yeunga:
@brianmajor A unit test has been added.

brianmajor:
@shinybrar - FYI: once merged, this PR will change how the units of RAM and CPU are shown in both the per-session and global stats.  Let us know if you want a heads up before the release so that we don't break your skaha client.

brianmajor (approved):

PR #421, opened 2022-12-21
CSD-3299: config change for session-desktop.yaml to isolate container selection

hjeeves:
Isolate selection of desktop instances from any desktop-app instances in a group of related desktop session containers. Related containers have the same sessionID, so the existing selector could return with a container that could not be connected to. 

Symptom on system: on connecting to a desktop session, there'd be intermittent 502 issues, or the arcade launch page would display three dots, or just return to the 'connect' button display with a websocket error.


kcasteels (approved):

kcasteels (approved):
Looks good.

kcasteels (approved):

kcasteels (approved):

ijiraq (approved):
Reviewed by Kevin

PR #431, opened 2023-01-25
Cadc 12189 update renew algorithm to accommodate different activeDeadlineseconds in skaha.sessionexpiry values

yeunga:
- for preliminary review
- implementation to be tested

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
+++ b/skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
@@ -68,11 +68,6 @@
 package org.opencadc.skaha;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import ca.nrc.cadc.util.StringUtil;
 
 public class K8SUtil {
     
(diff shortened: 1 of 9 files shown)


skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
```diff
@@ -68,6 +68,11 @@
 package org.opencadc.skaha;
 
 import java.io.IOException;
+import java.util.HashMap;
```
hjeeves: Are these imports needed?

hjeeves: Other than those imports, it seems good.

hjeeves (changes requested):
Sounds like just the k8sUtil imports should be walked back.

skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
yeunga: Removed unnecessary Imports.

hjeeves (approved):
Looks good! Thanks for updating the integration tests.

brianmajor (approved):

Issue #430, opened 2023-01-24
skaha-pluto build fails


brianmajor:
Step 8 of the skaha-pluto build is currently failing.

https://github.com/opencadc/science-platform/blob/8cc7970f7af4ab8a46f88d816afae082d6994b3e/containers/session-containers/skaha-pluto/Dockerfile#L53

I haven't dug into it yet, but before doing that I was wondering any anyone (especially @sefffal ) had some quick insight.  The error is:

```
#12 152.6 ERROR: failed process: Process(`/julia-1.8.0-beta3/bin/julia --color=auto --startup-file=no --cpu-target=native -O3 --sysimage=/julia-1.8.0-beta3/lib/julia/sys.so --project=/opt/julia/environments/v1.8 --output-o=/tmp/jl_HvjqbPslUl.o /tmp/jl_X3Ur7Anhde`, ProcessSignaled(9)) [0]
#12 155.0
#12 155.0 Stacktrace:
#12 155.0  [1] pipeline_error
#12 155.0    @ ./process.jl:561 [inlined]
#12 155.0  [2] run(::Cmd; wait::Bool)
#12 155.1    @ Base ./process.jl:476
#12 155.1  [3] run(::Cmd)
#12 155.1    @ Base ./process.jl:474
#12 155.1  [4] (::PackageCompiler.var"#16#18")()
#12 155.1    @ PackageCompiler /opt/julia/packages/PackageCompiler/kh9hl/ext/TerminalSpinners.jl:157
#12 155.1  [5] spin(f::PackageCompiler.var"#16#18", s::PackageCompiler.TerminalSpinners.Spinner{Base.PipeEndpoint})
#12 155.1    @ PackageCompiler.TerminalSpinners /opt/julia/packages/PackageCompiler/kh9hl/ext/TerminalSpinners.jl:164
#12 155.1  [6] macro expansion
#12 155.1    @ /opt/julia/packages/PackageCompiler/kh9hl/ext/TerminalSpinners.jl:157 [inlined]
#12 155.1  [7] create_sysimg_object_file(object_file::String, packages::Vector{String}, packages_sysimg::Set{Base.PkgId}; project::String, base_sysimage::String, precompile_execution_file::Vector{String}, precompile_statements_file::Vector{String}, cpu_target::String, script::Nothing, sysimage_build_args::Cmd, extra_precompiles::String, incremental::Bool)
[...]

PR #432, opened 2023-01-25
Bump Julia version in pluto image

sefffal:
Should address #430.
Ideally we could change this to take the Julia version in as a variable.


brianmajor (approved):

brianmajor:
Many thanks @sefffal !

PR #433, opened 2023-01-25
Run skaha in a replica set of 3 instances.

brianmajor:


brianmajor:
@kcasteels - FYI - This is how the replica set is adjusted.

PR #434, opened 2023-01-27
OpenCADC harbor deployment files

kcasteels:


deployment/k8s-config/harbor/opencadc/pvc-mount.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: batch/v1
```
brianmajor: Is this file for debugging the PVC?  Should it be in here?  If you think it should then its use should be documented.

brianmajor (changes requested):
Just a few points @kcasteels :
1. Could you provide a README.md with brief install instructions?
2. Since there is now an `opencadc` folder, could you create a new folder for images.canfar.net (not what that would be) and move the existing files in there?  They should be on the same level for organization purposes.
3. And just a question about whether we need that PVC yaml file...

brianmajor (changes requested):
Thanks for the reorganization, but I think you missed my comments about a README.md and there's a question about the pvc-mount.yaml file.

deployment/k8s-config/harbor/opencadc/pvc-mount.yaml
kcasteels: I've moved this file to deployment/k8s-config/scripts directory as it is a debugging tool.

brianmajor (approved):

PR #439, opened 2023-02-08
Cadc 12192 Update session deletion to be more tolerant of 'not found' errors.

yeunga:


skaha/src/main/java/org/opencadc/skaha/session/DeleteAction.java
```diff
@@ -137,27 +138,37 @@ public void doAction() throws Exception {
...
+                "kubectl", "delete", "--namespace", k8sNamespace, type, name};
             execute(cmd);
+        } catch (IOException ex) {
+            String msg = ex.getMessage();
+            if (msg.contains("NotFound") || msg.contains("not found")) {
+                // fail to find the object to delete, log a warning and continue
+                log.warn(msg);
+            } else {
+                // some other error, propagate the exception
+                throw ex;
```
brianmajor: Instead of propagating the other types of exceptions, just do a log.warn (like you do above) instead.  The matching of exception text is probably not reliable between k8s upgrades etc anyhow.

brianmajor (changes requested):
Just the one change Alinga

skaha/src/main/java/org/opencadc/skaha/session/DeleteAction.java
yeunga: Done.

skaha/src/main/java/org/opencadc/skaha/session/DeleteAction.java
```diff
@@ -137,27 +138,31 @@ public void doAction() throws Exception {
...
+            delete(k8sNamespace, "service", serviceName);
+        }
+    }
+    
+    private void delete(String k8sNamespace, String type, String name) throws InterruptedException, IOException {
+        try {
+            String[] cmd = new String[] {
+                "kubectl", "delete", "--namespace", k8sNamespace, type, name};
             execute(cmd);
+        } catch (IOException ex) {
```
brianmajor: You may as well catch 'Exception' in stead of just 'IOException' to make sure we get catch the runtime ones as well.

brianmajor (changes requested):

brianmajor (approved):

skaha/src/main/java/org/opencadc/skaha/session/DeleteAction.java
yeunga: Sure. This changes the initial intent of the ticket, but works.
Done.

PR #440, opened 2023-02-09
CADC-12073 Synchronized injectProxyCert(...) to avoid collision

yeunga:
Launched new desktop and notebook sessions and verified that the cadcproxy.pem file was updated each time.

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
+++ b/skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
@@ -237,7 +237,7 @@ public abstract class SessionAction extends SkahaAction {
         return "https://" + host + "/session/contrib/" + sessionID + "/";
     }
     
-    protected synchronized void injectProxyCert(final Subject subject, String userid, String posixID)
+    protected void injectProxyCert(final Subject subject, String userid, String posixID)
             throws PrivilegedActionException, IOException, InterruptedException {
         
         // creating cert home dir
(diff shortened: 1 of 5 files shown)


skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -237,7 +237,7 @@ public static String getContributedURL(String host, String sessionID) throws Mal
         return "https://" + host + "/session/contrib/" + sessionID + "/";
     }
     
-    protected void injectProxyCert(final Subject subject, String userid, String posixID)
```
brianmajor: I think having the monitor on this method will cause proxy injection to be synchronized for all users, not just the one launching a new session.  Plus, since ~/.ssl is on arc they could be doing things in there without going through skaha.  I would suggest making the injection atomic by creating the file in /tmp/ and then moving it to ~/.ssl/cadcproxy.pem.

brianmajor (changes requested):

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: Refactored SessionAction.injectProxyCert(...). Please take a look. Thanks.

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -291,6 +283,33 @@ protected String getImageName(String image) {
...
+        writer.close();
+        
+        // update file permissions
+        String[] chown = new String[] {"chown", posixID + ":" + posixID, tmpFileName};
+        execute(chown);
+//        String[] chown = new String[] {"chown", "-R", "guest:guest", "/home/" + userid + "/.ssl"};
+//        execute(chown);
+        
+        // inject file
+        String[] injectCert = new String[] {"cp",  "-rp", tmpFileName, homedir + "/" + userid + "/.ssl/cadcproxy.pem"};
```
brianmajor: @yeunga From what I've read (though I'm not an expert), the `cp` command is not atomic, but the `mv` command is...

brianmajor (changes requested):

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
yeunga: Sure, I can change from 'cp' to 'mv' and remove the clean up code.

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -293,6 +285,29 @@ protected String getImageName(String image) {
 
     }
     
+    protected synchronized void injectFile(String data, String posixID, String userid) throws IOException, InterruptedException {
```
brianmajor: Looks good Alinga.  This `synchronized` modifier should be removed now though.

brianmajor (changes requested):

brianmajor (approved):

PR #444, opened 2023-02-16
Cadc 12267 Fixed attaching a desktop app to a desktop.

yeunga:
Also updated GetAction.getCPUCores(...) to handle millicore unit.

Added ACIdentityManager to skaha catalina.properties. 
Added ACIdentityManager to arc catalina.properties. (not tested)

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
+++ b/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
@@ -253,7 +253,7 @@ public class PostAction extends SessionAction {
                     throw new IllegalArgumentException("Missing parameter 'image'");
                 }
                 
-                attachSoftware(image, params);
+                attachDesktopApp(image, params);
                 
             } else {
                 throw new UnsupportedOperationException("Cannot modify an existing app.");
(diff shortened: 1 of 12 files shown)


hjeeves:
Verified in test that Platform Load panel is indeed loading again (millicore unit issue)  and can open >1 terminal (desktop-app) within a desktop session.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -572,7 +576,8 @@ public void attachSoftware(String image, List<String> params) throws Exception {
                     log.debug("part: " + part);
                 }
             }
-            if (parts.length > 1 && parts[1].trim().equals("<none>")) {
+            if (parts.length > 1 && parts[1].trim().equals(NONE) && 
+                    SESSION_TYPE_DESKTOP.equals(parts[2])) {
                 targetIP = parts[0].trim();
```
brianmajor, line 581: I think it would have been better to add the sessionType to the selector (rather than get multiple results back).  Maybe in another PR you could try that change.  Otherwise, looks good.  Thanks.

brianmajor (approved):

PR #451, opened 2023-04-12
Fix for #430

sefffal:
This PR addresses issue number #430 , pluto failing to launch.

I removed the step creating a "system image". This was failing, but is now somewhat redundant on Julia 1.9 (which incidentally I bumped from beta3 to rc2).

### DIFF
--- a/containers/session-containers/skaha-pluto/Dockerfile
+++ b/containers/session-containers/skaha-pluto/Dockerfile
@@ -6,9 +6,9 @@ RUN yum makecache -y && \
     yum clean all
 
 WORKDIR /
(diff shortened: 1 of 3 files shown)


brianmajor (changes requested):
Thanks @sefffal ! - Seems to be working.  In VERSION, perhaps change the first tag from `0.2` to `1.9rc2` to reflect the version of julia?

sefffal:
Done. It would be nice if that version could be used to control what gets built into the Dockerfile.

brianmajor:
Yep, I agree.
Thanks again.  I've pushed it to harbor so is now available publicly under the 'contributed' type.
Cheers.

brianmajor (approved):

sefffal:
Thanks!

PR #479, opened 2023-07-04
SP-3544 Rajat/Saurabh Added helm chart for skaha

rob-tw:


at88mph (approved):

at88mph:
Much of the items in the `catalina.properties` file will need to come from the `values.yaml` file during a Helm install.  I'll merge this now and then we can just iterate on items like that.

PR #481, opened 2023-07-11
RED-11 | Abhishek/Hari | Added IAM support for IAM

abhishekghoshhh:
1. Added development changes in SkahaAction.java
2. Added integration test for /skaha/v0/context api
3. Added .gitignore


integration-test/configs/iam/keystore.jwks
```diff
@@ -0,0 +1,16 @@
...
+      "e": "AQAB",
+      "kid": "rsa1",
+      "qi": "bN1wnq_0VkJlECMGPmeRFdZCX2LgAMrgwbJpysRw5J04vO9YsVmAcB_4xqoDjDUg7koioAp3IOhMGjOJWpYzCqWzsaA_84kX4WKGsr2xz6oaFSgt1FvtBY4GqEeZj8RG0LMtEtKSyjHGieA0hd4TUcqSdNZ4osT98Bfd7z3peYc",
+      "dp": "YfyaxI2IRHyXavm9M-hAIWH2JNOD5gGJU5p8_cnw9NHlRuZNZJF16p5sEAxh6tn1MtsvsTxrMx_RvjqEp2IsEXaaOcZN0v7zhwlfcxMZT-TC-eQkH7rLg4Wz_dOVytt4FpFWPpySuloGjusXKLNhBeDi31dMo5SeYQvpj0k8iek",
+      "dq": "9_lhyLPNdohmxqwE5kkA7L23NbPJ-svmavWBwo3HMlCiLkQoeCEx8EzebsCux9-wfKSuSqfHrtCALU15QxUR6x2SdeRvVY17cGHm3kNTA_4j8cbBYdccjXSksitzZ-wOfvVDjxcqST2llkm8NjoO18Siv0-F4SXKLG-c5CaE9w",
+      "n": "4GRvJuFantVV3JdjwQOAkfREnwUFp2znRBTOIJhPamyH4gf4YlI5PQT79415NV4_HrWYzgooH5AK6-7WE-TLLGEAVK5vdk4vv79bG7ukvjvBPxAjEhQn6-Amln88iXtvicEGbh--3CKbQj1jryVU5aWM6jzweaabFSeCILVEd6ZT7ofXaAqan9eLzU5IEtTPy5MfrrOvWw5Q7D2yzMqc5LksmaQSw8XtmhA8gnENnIqjAMmPtRltf93wjtmiamgVENOVPdN-93Nd5w-pnMwEyoO6Q9JqXxV6lD6qBRxI7_5t4_vmVxcbbxcZbSAMoHqA2pbSMJ4Jcw-27Hct9jesLQ"
+    }
+  ]
+}
```
brianmajor: Does the inclusion of these keys compromise access to IAM at all?  The opencadc organization is fully public.

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
```diff
@@ -180,6 +192,29 @@ protected void initRequest() throws Exception {
         LocalAuthority localAuthority = new LocalAuthority();
         URI gmsSearchURI = localAuthority.getServiceURI("ivo://ivoa.net/std/GMS#search-0.1");
 
+        log.debug("using iam to fetch group details");
+        cacheGroupsWithIAM(subject, gmsSearchURI);
```
brianmajor: You can delete the `cacheGroupsWithIAM` method and just put the code right in here.

The cacheGroupsWithForLDAP method can be deleted.

brianmajor (changes requested):
Looking good Abhishek and Hari.  A few comments have been added, but also:

We can remove the concept of 'admin' all together.  Could you please delete the boolean `protected boolean adminUser = false` from SkahaAction.java, and then remove whatever other functionality it was used for?  I believe it allowed for the retrieval of images from a non-harbor registry.  That ability should be removed for now until we get to the story about integrating with other docker registries.

integration-test/configs/iam/keystore.jwks
abhishekghoshhh: This keystore.jwks is built locally and is used by the **local iam** only in docker compose which is required for the integration test. 
I can remove this from the pull request but then everytime before running the integration test we have to create keystore.jwks.
For refference : https://indigo-iam.github.io/v/v1.8.0/docs/getting-started/jwk/

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
abhishekghoshhh: I have deleted this cacheGroupsWithForLDAP method and also move the contents of cacheGroupsWithIAM function in the caller function

integration-test/configs/iam/keystore.jwks
brianmajor: Okay, thanks.  Seems like it's okay to keep it.

deployment/helm/skaha/skaha-config/launch-carta.yaml
```diff
@@ -31,6 +31,13 @@ spec:
         supplementalGroups: [${skaha.supgroups}]
       priorityClassName: uber-user-preempt-medium
       hostname: "${software.hostname}"
+      initContainers:
+      - name: setup-access-token
+        image: busybox
+        command: ["sh", "-c", "echo '${access_token}' > /etc/tokens/access-token"]
```
brianmajor: @andamian - Could you please advise here?  Where do/should client tools look for tokens to be present on a machine?  Something like ~/.ssl/cadcproxy.pem, but for tokens, and more general and standard for the python community.

deployment/helm/skaha/skaha-config/launch-contributed.yaml
```diff
@@ -31,6 +31,13 @@ spec:
         supplementalGroups: [${skaha.supgroups}]
       priorityClassName: uber-user-preempt-medium
       hostname: "${software.hostname}"
+      initContainers:
+      - name: setup-access-token
+        image: busybox
+        command: ["sh", "-c", "echo '${access_token}' > /etc/tokens/access-token"]
```
brianmajor: Could you make this variable match the others in terms of convention?  `${skaha.accesstoken}` perhaps?

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
```diff
@@ -98,8 +96,14 @@
...
 import org.json.JSONArray;
 import org.json.JSONObject;
+import org.opencadc.auth.StandardIdentityManager;
 import org.opencadc.gms.GroupURI;
+import org.opencadc.gms.IvoaGroupClient;
 import org.opencadc.skaha.image.Image;
+import org.springframework.util.StringUtils;
+
+import static java.util.stream.Collectors.toList;
+import static org.springframework.util.CollectionUtils.isEmpty;
```
brianmajor: It would be nice it we didn't have to bring in the springframework jar file to support this command.  But, I don't see a reference to it in build.grade?

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
```diff
@@ -180,40 +182,21 @@ protected void initRequest() throws Exception {
...
-        }
-
-        if (!memberships.contains(skahaUsersGroupObj)) {
+        log.debug("using iam to fetch group details");
+        IvoaGroupClient ivoaGroupClient = new IvoaGroupClient();
+        GroupURI skahaUsersGroupUri = new GroupURI(URI.create(skahaUsersGroup));
+        Set<GroupURI> skahaUsersGroupUriSet = ivoaGroupClient.getMemberships(gmsSearchURI);
+        log.debug("skahaUsersGroupUriSet is " + skahaUsersGroupUriSet);
+        if (!skahaUsersGroupUriSet.contains(skahaUsersGroupUri)) {
+            log.info("user is not a member of skaha user group ");
```
brianmajor: In openCADC software, the INFO level of log is supposed to only be used by our RestAction framework.  So, could you please change the log.info lines to be log.debug instead?  We normally run services in INFO level so don't want too much noise in there.

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
```diff
@@ -111,6 +115,7 @@ public abstract class SkahaAction extends RestAction {
     public static final String SESSION_TYPE_CONTRIB = "contributed";
     public static final String SESSION_TYPE_HEADLESS = "headless";
     public static final String TYPE_DESKTOP_APP = "desktop-app";
+    public static final String TRUE = "true";
```
brianmajor: Is this being used?  If so, I think Boolean.TRUE.toString() (or something like that) is the same.

brianmajor (changes requested):
Nearly there... :)

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
abhishekghoshhh: It is not used currently. I have removed it from the code.

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
abhishekghoshhh: I have changed that specific log to debug level

skaha/src/main/java/org/opencadc/skaha/SkahaAction.java
abhishekghoshhh: It is shown in the import suggestion, That's why I have used.
Currently I have removed it and added a method to check if the collection is empty or not

deployment/helm/skaha/skaha-config/launch-contributed.yaml
abhishekghoshhh: I have updated the keys as per the convension.

deployment/helm/skaha/skaha-config/launch-carta.yaml
andamian: I'm still a newbie when it comes to the use of tokens. That being said, the tokens must be stored in user space to avoid collision with other users. I don't know of any "standard" cache directory (such as `~/.ssl/`). Maybe `~/.tokens`?

Does IAM support refresh tokens? Access tokens are typically short lived tokens that might be stored in memory for the duration of the cli application whereas refresh tokens are longer lived tokens that are used to retrieve access tokens and can be stored locally in a file.

brianmajor (commented):
An issue has just occurred to me: a deployer can configure skaha to use different identity managers.  In this case the assumption here is that the StandardIdentityManager is configured and that the user with authenticate with a Bearer token.  However, we must allow for the case where a different identity manager is used and we may not have an access token.  Similarly, the existing code (before this PR) must now allow for the StandardIdentityManager will be used, which means that a proxy certificate is not available and thus cannot be injected.

The changes to the launch-*.yaml files, for example, would only work if we have a token by way of using a StandardIdentityManager configuration.  (The initContainer cannot run in this case.)  So, we need a design that will work when we use either of those Identity Managers.

PR #491, opened 2023-08-04
Update local authority.

at88mph:


brianmajor (approved):

brianmajor:
Should be in cadc-registry.properties instead.

PR #495, opened 2023-08-07
RED-16 | Abhishek/Hari | Cleaning up old init container configs

abhishekghoshhh:


skaha/src/main/java/org/opencadc/skaha/FileParser.java
```diff
@@ -0,0 +1,98 @@
+package org.opencadc.skaha;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class FileParser {
```
at88mph: Is there a unit test for this?  Where is this used?  I think this was just to load from the text file for the stubbed out users, correct?

at88mph (approved):

PR #500, opened 2023-08-18
CADC-12724 Added 'enableServiceLinks=false' to all launch*.yaml files.

yeunga:
Since only deployment files were changed, no VERSION file was updated.


brianmajor (changes requested):
- `cadc-registry.properties` in base/skaha/config points to rc-ws, but should be prod instead
- The overlay versions of `cadc-registry.properties` are missing.  Those should point to rc.ws.

Lastly, did you check and see if the environment variables from other sessions are gone with the addition of `enableServiceLinks: false`?

brianmajor:
Oh, never mind my question, I see your comment in Jira that the setting works as expected.

brianmajor (approved):

PR #514, opened 2023-09-22
Sp 3544

abhishekghoshhh:
Creating a pull request for code to inject POSIX user and group information into a job pod. 
This implementation does not require any container to run as root.


at88mph:
This has all been moved into a REST API at
https://github.com/opencadc/ac/tree/master/posix-mapper

deployment/helm/skaha/skaha-config/launch-notebook.yaml
```diff
@@ -25,12 +25,29 @@ spec:
...
+        - ${user.posix.entry}
+        - ${user.group.entry}
+        volumeMounts:
+        - mountPath: "/etc-passwd"
+          name: etc-passwd
+        - mountPath: "/etc-group"
+          name: etc-group
+        - mountPath: "/skaha-system"
+          name: start-jupyterlab
+        securityContext:
```
rptaylor, line 48: The container securityContext is redundant and could be removed since the securityContext of the pod applies by default, see `kubectl explain pod.spec.containers.securityContext`

deployment/helm/skaha/skaha-config/launch-notebook.yaml
```diff
@@ -25,12 +25,29 @@ spec:
...
       hostname: "${software.hostname}"
+      initContainers:
+      - name: init-notebook
+        image: ${software.imageid}
+        command: ["/skaha-system/init-jupyterlab.sh"]
+        args:
+        - ${user.posix.entry}
+        - ${user.group.entry}
+        volumeMounts:
+        - mountPath: "/etc-passwd"
```
rptaylor, line 42: Would be simpler to use a single volume for both /etc/passwd and /etc/group

deployment/helm/skaha/skaha-config/launch-notebook.yaml
```diff
@@ -86,6 +102,12 @@ spec:
           protocol: TCP
           name: notebook-port
         volumeMounts:
+        - mountPath: "/etc/passwd"
```
rptaylor, line 105: In my testing I could not get a subpath mount of an emptydir volume to work, but I found documentation references suggesting it should work, so I am curious what specific issue might have been preventing this in my tests or if you were able to test and confirm that it works, if it is only supported in some k8s versions, etc.

rptaylor: Also the volume mounts in the run container must be read-only.

deployment/helm/skaha/skaha-config/launch-notebook.yaml
```diff
@@ -104,3 +126,7 @@ spec:
           claimName: skaha-workload-cavern-pvc
       - name: scratch-dir
         emptyDir: {}
+      - name: etc-group
+        emptyDir: {}
```
rptaylor, line 130: Setting a small size limit e.g. 5 MB can be a good protective measure.

deployment/helm/skaha/start-jupyterlab/init-jupyterlab.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# One argument expected: the skaha session ID, which makes up part of the connect url
+
```
rptaylor, line 4: As this script is executed by the initcontainer which runs the user-defined image, just pointing out a security consideration that /bin/bash, cp etc could be malicious executables containing arbitrary code. 
Of course users can run their own images anyway which could contain anything (and must be fully unprivileged), so the only consideration in scope here is the potential impact of any such malicious code taking part in preparation of the passwd and group files.
(If the passwd and group files were just configmap subpath mounts this would not be an issue.)

PR #539, opened 2023-10-12
Pi20

abhishekghoshhh:
**changes** : 
1. skaha TLD
2. min uid
3. passing posix info directly into yaml
4. container resource increase
5. posix mapper unused code removal


deployment/helm/posix-mapper/templates/posix-mapper-tomcat-expose.yaml
```diff
@@ -10,6 +10,10 @@ spec:
   - port: 8080
     name: http-connection
     protocol: TCP
+  - port: 80
+    targetPort: 8080
+    name: http-connection-automatic
+    protocol: TCP
```
at88mph, line 16: I guess this won't hurt, but we do install Traefik in front of these services which should proxy to the 8080 port.  Is this just to enable Port 80 without a proxy service?

at88mph (changes requested):

deployment/helm/posix-mapper/values.yaml
at88mph, line 1: Please keep the `minUID` value, that's good.  Please restore the rest of the values though.

at88mph (changes requested):

at88mph (approved):

PR #541, opened 2023-10-17
CADC-12826 Updated deployment to make it easier to add an icon to the…

yeunga:
… skaha-desktop and be able to click on the icon to start an existing desktop-app container


at88mph (dismissed):

at88mph:
Could you also add a `VERSION` increment, @yeunga?

yeunga:
Currently we don't have versioning on the deployment files. The changes only affect deployment files.

brianmajor:
Some comments:
- As mentioned in slack, I don't like how this functionality depends on things (such as the icons) being available in the desktop container.  Could we put the icons in as a configmap and mount them instead?
- Please add the instructions to the README.md in skaha-workload
- There is a more general issue: imagine a deployer (other that us) that wants to have different icons in the desktop than what we have.  It wouldn't be possible for them to commit those changes back to git because it would then change our desktop.  So, the config of the icons must be outside of git, but can still be part of the deployment.  With helm this is achieved through the values.yaml files.  In kustomize, I guess we just look for a configmap with that file name and use it if available.  In both cases, you need a default behaviour, which can be behaviour we currently have configured.

yeunga:
Sure. I can look into making the desktop-app icons configurable. I wasn't sure if this was feasible. That's the next thing I wanted to do anyways.

brianmajor (approved):

PR #567, opened 2023-11-09
CADC-12887 Refactored build-menu.sh init to clean up the code 

yeunga:
We now delete and recreate ~/.local/skaha if already exists instead of just emptying the directory. Also updated code to use '(PROJECT)' instead of '(NAME)'.


at88mph (approved):

brianmajor:
@yeunga Has this been shown to address the menu bug?  Merging to master means it is ready for release.

PR #579, opened 2023-12-06
Removed code which added user to the skaha-users group

kcasteels:
This commit removes code which added users to the skaha-users group, but this is redundant because the users should already be a member of skaha-users in order to gain access to the portal in the first place.


brianmajor (approved):
Good catch Kevin!

PR #595, opened 2023-12-20
Pi20

abhishekghoshhh:
Before deployment add private and public key pair **(skaha-private.pem + skaha-public.pem)** and deployment/helm/skaha/token-tool-secret folder.


at88mph:
Thanks @abhishekghoshhh .  I'm putting this into a separate branch as we're trying to merge everything into `master`.

PR #597, opened 2024-01-10
Helm install and SRCNet compatibility

at88mph:


brianmajor (commented):
Hi @at88mph - Just trying to understand Pablo's storage problem.  Are there indeed two PVCs for cavern storage, one for each namespace?  Pretty sure they used to use the same one.  Is there a technical limitation there with helm?

Also, I see that in skaha-tomcat-deployment.yaml there is config for adding additional volumes, but I don't see why they would be in there and not in the user containers (eg launch-notebook.yaml).

at88mph:
They used to be declared repetitively dynamically in each configuration file.  However, `PersistentVolumeClaim`s have a namespace, and there is a one-to-one relationship with the `PersistentVolume`s and the `PersistentVolumeClaim`s, which means there are two `PersistentVolume`s with the same configuration, and two `PersistentVolumeClaim`s, one for each volume, and in their respective Namespaces.

If the sessions and Skaha used Cavern in a client-server way, then there would be a single `PersistentVolume` and `PersistentVolumeClaim` just for Cavern.  Mounting our storage presents this redundant complexity unfortunately.

brianmajor:
Think we'll have to discuss this in person--too complex over github!

What about the second part of my question though, about the additional volume configuration support being in skaha rather than in the sessions?

brianmajor (approved):

PR #602, opened 2024-01-30
Update the RAM selections.

at88mph:


brianmajor (commented):
Might as well make the assocated change (whatever it ends up being) on the helm side too.

sfabbro (commented):
Looks good. I would add a few RAM slots around the threshold of 32GB where GPU users can select the V100 vs A100 nodes.
So would add: 26, 30, 36, 44 to the RAM choices.

fraserw:
I wonder if adding a 0.5 GB ram option is possible? I have the rare headless task for which that value would be fine.

at88mph:
I've added extra options for @sfabbro suggestions.  The current backend supports integers only, @fraserw, but if there is enough interest we could update the configuration code.

fraserw:
Ah ok NVM. Mine was a pretty edge case.

at88mph:
This need to get merged before closing.

sfabbro (approved):

PR #606, opened 2024-02-14
Changes to session priorities

brianmajor:
For [CADC-13195](https://cadc-ccda.atlassian.net/browse/CADC-13195):
- Restoration of preemptive behaviour for interactive sessions
- Addition of high priority headless group and configurable priority class
- Fix lost regular headless group setting

[CADC-13195]: https://cadc-ccda.atlassian.net/browse/CADC-13195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
+++ b/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
@@ -976,8 +976,12 @@ public class PostAction extends SessionAction {
         if (skahaPriorityHeadlessGroup == null) {
             return "";
         }
+        if (skahaHeadlessPriortyClass == null) {
+            log.warn("headlessPriorityGroup set but headlessPriorityClass not set");
+            return "";
+        }
         if (priorityHeadlessUser) {
-            return "priorityClassName: uber-user-preempt-medium";
+            return "priorityClassName: " + skahaHeadlessPriortyClass;
         } else {
             return "";
         }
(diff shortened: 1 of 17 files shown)


at88mph:
Reviewing this now.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -978,5 +971,20 @@ private String getGPUScheduling(Integer gpus) {
         }
         return sb.toString();
     }
+    
+    private String getHeadlessPriority() {
+        if (skahaPriorityHeadlessGroup == null) {
+            return "";
+        }
+        if (skahaHeadlessPriortyClass == null) {
+            log.warn("headlessPriorityGroup set but headlessPriorityClass not set");
```
at88mph, line 980: Should this actually throw an Exception or is this an acceptable incomplete configuration?

at88mph (changes requested):

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
brianmajor: I figure that if deployers don't need this headless priority function then they shouldn't need to configure it.  So, left it as just a warning.

at88mph (approved):

PR #635, opened 2024-05-10
SSHD deploy

at88mph:
- Deploy SSHD service
- Include key generation in POSIX Mapper for API Key access
- Include extraconfig change to allow deployment of CADC specific resources, such as RsaSignatureKey.pub
- Cavern API updates
- Added loggingGroups to APIs to allow setting /logControl levels
- Cleanup and minor fixes, such as reporting required values that are missing


deployment/helm/sshd/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+To faciliate password-less logins, please upload a Public Key to your Home directory in the `.ssh/authorized_key` file.  The easiest way to do that
+is to create a local `authorized_keys` file, or you may use your existing one if it contains your Public Key already.  If you're creating a new one, simply put the contents
+of your Public Key into the `authorized_keys` file:
+
+```sh
+$ cat ~/.ssh/id_rsa.pub > authorized_keys
+```
+
+Then create a folder in the called `.ssh` in your `/home/<username>` folder.  It's quickest to use the Storage UI to do that.  Assuming your storage is called `cavern` and your username is `exampleuser`:
```
brianmajor: We'll need to document how to change the permissions of ~/.ssh and ~/.ssh/authorized_keys (chmod or vchmod), so perhaps an alternative access method to cavern is preferable.  Probably a live session and user can just copy/paste the key value into authorized_keys?

brianmajor: Actually, all this about uploading a public key should go into the user-facing documentation and not the deployer documentation.

brianmajor (changes requested):
Just the documentation suggestion

deployment/helm/sshd/README.md
at88mph: Sure.  Maybe a new page on the CANFAR site Documentation then?

brianmajor: The science-containers.git repo is the place for user-facing documentation.  It has ci/cd built-in which publishes the docs to https://www.opencadc.org/science-containers/

at88mph: Done.
https://github.com/opencadc/science-containers/pull/104

at88mph:
A quick note about the `authorized_keys` file's permissions; all the testing with uploading through the UI have shown that it works as-is.  Doing it through a Notebook, for example, might have a different outcome though.  The main documentation is for uploading it with the UI.

deployment/helm/sshd/README.md
```diff
@@ -0,0 +1,79 @@
...
+STATUS: deployed
+REVISION: 1
+TEST SUITE: None
+```
+
+## Verification
+
+From a terminal:
+```sh
+$ ssh -p 64022 example.org
```
brianmajor, line 78: - Is 64022 only applicable to our deployment?  I could be wrong but it seems like the ingress and service are listening on 2222.
- Does the userid need to be included in the call?

deployment/helm/sshd/templates/sshd-deployment.yaml
```diff
@@ -0,0 +1,181 @@
+{{- $image := "linuxserver/openssh-server:latest" -}}
```
brianmajor, line 1: I wonder if this should be fixed to a certain version so that we are not surprised by an update that is incompatible with our config or our deployment...

brianmajor (changes requested):

deployment/helm/sshd/README.md
at88mph: The 64022 is the default that is opened by Traefik.  This is just part of the example, which is derived from our deployment.  Port 2222 is served from the Docker Image, and mapped through the Traefik endpoint that's serving 64022.  It assumes the current user is the one to login with.

deployment/helm/sshd/templates/sshd-deployment.yaml
at88mph: That may be a good idea, sure.  I'll find the latest one.

PR #647, opened 2024-07-03
Posix map fix service call

at88mph:
Backup original `/etc/passwd` and `/etc/group` from the User Session Image.
Merge original files with POSIX Mapper calls
- Uses `images.opencadc.org/library/cadc-tomcat:1` image as it contains all of the necessary packages to make call to POSIX Mapper.  Does not start Tomcat, just uses the installed packages.  We can revisit this later, likely with a custom image.


skaha/src/main/java/org/opencadc/skaha/utils/PosixHelper.java
```diff
@@ -17,15 +43,95 @@ public static String uidMapping(final PosixPrincipal posixPrincipal) {
...
-     * @param sessionId The session identifier to mark the secret.
-     * @return String secret name, never null.
-     */
-    public static String getPosixMapperSecretName(String sessionId) {
-        return "posix-mapping-" + sessionId;
+    static Date currentDateUTC() {
+        return Calendar.getInstance(DateUtil.UTC).getTime();
+    }
+
+    static boolean posixMappingSecretRequiresUpdate() throws Exception {
```
brianmajor: I guess these two methods and the executeInShell method aren't needed with this new approach

brianmajor (changes requested):

skaha/src/main/java/org/opencadc/skaha/utils/PosixHelper.java
at88mph: The entire class can go, actually.

brianmajor (approved):

PR #652, opened 2024-07-09
Posix map fix

at88mph:
Add kubectl client 1.25 to reflect current cluster minimum (after upgrade).
Correct how the `extraHosts` property is used.
Each `launch` script now has a single init container.


at88mph:
Closing to replace with #666 .

PR #653, opened 2024-07-10
Add extraHosts support to posix mapper, following local convention

daikema:
When attempting to setup the POSIX mapper I was having errors reaching my local registry and discovered that that was due to the hostname I was using resolving to the local loopback address and thus wasn't functioning from within a pod.

Most of the other charts had support for adding host aliases, so I've added the same for the posix mapper helm chart following the conventions from the other charts in the repo.

### DIFF
--- a/deployment/helm/posix-mapper/templates/posix-mapper-tomcat-deployment.yaml
+++ b/deployment/helm/posix-mapper/templates/posix-mapper-tomcat-deployment.yaml
@@ -45,6 +45,12 @@ spec:
         {{- with .Values.deployment.posixMapper.extraVolumeMounts }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
+    {{- range $extraHost := .Values.deployment.extraHosts }}
+      hostAliases:
+        - ip: {{ $extraHost.ip }}
+          hostnames:
+            - {{ $extraHost.hostname }}
+    {{- end }}
       volumes:
       - name: config-volume
         configMap:
--- a/deployment/helm/README.md
+++ b/deployment/helm/README.md
@@ -149,6 +149,18 @@ deployment:
     #     defaultMode: 420
     #     secretName: posix-manager-cacert-secret
 
+  # Specify extra hostnames that will be added to the Pod's /etc/hosts file.  Note that this is in the
+  # deployment object, not the posixMapper one.
+  #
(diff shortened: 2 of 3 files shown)


deployment/helm/posix-mapper/templates/posix-mapper-tomcat-deployment.yaml
```diff
@@ -45,6 +45,12 @@ spec:
         {{- with .Values.deployment.posixMapper.extraVolumeMounts }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
+    {{- range $extraHost := .Values.deployment.extraHosts }}
+      hostAliases:
+        - ip: {{ $extraHost.ip }}
+          hostnames:
+            - {{ $extraHost.hostname }}
+    {{- end }}
```
at88mph, line 53: There's actually a bug with this that will be fixed in the next release in the other deployments, so we'll need the same fix here.  The `range` will add multiple `hostAliases` stanzas, but only one is allowed, with an array of entries.  So the `range` needs to be _inside_ the `hostAliases` entry, but only if some were specified.  It should look like:

```yaml
{{- with .Values.deployment.extraHosts }}
      hostAliases:
{{- range $extraHost := . }}
        - ip: {{ $extraHost.ip }}
          hostnames:
            - {{ $extraHost.hostname }}
{{- end }}            
{{- end }}
```

at88mph (changes requested):

brianmajor:
The main issue with this approach is that skaha assumes knowledge and makes use of the cavern mount--something we're trying to get away from.  

The currently released version, though not portable, seems to be performing well.  I think something similar to that design is ideal.  Having a container that does the merge be based on cadc-java would be an improvement.

daikema:
@at88mph are you able to add that change you suggested here?  I'm currently on leave.

at88mph:
> The main issue with this approach is that skaha assumes knowledge and makes use of the cavern mount--something we're trying to get away from.
> 
> The currently released version, though not portable, seems to be performing well. I think something similar to that design is ideal. Having a container that does the merge be based on cadc-java would be an improvement.

I think you meant for this to go into #652, @brianmajor?

at88mph:
Sure thing, @daikema .

at88mph (approved):

PR #664, opened 2024-07-18
Explicit/refactored security contexts for containers

rptaylor:
```
kubectl explain pod.spec.securityContext
kubectl explain pod.spec.containers.securityContext
```
Things that can only be in pod securityContext:
	fsGroup, fsGroupChangePolicy
	supplementalGroups
	sysctls
Things that can only be in container securityContext:
	allowPrivilegeEscalation
	capabilities			
	privileged			
	procMount
	readOnlyRootFilesystem
Everything else can be in either, but pod-level properties apply to all containers by default unless overridden by the container securityContext.

From what I saw every pod spec of every job already had
```
      securityContext:
        runAsUser: ${skaha.posixid}
        runAsGroup: ${skaha.posixid}
        fsGroup: ${skaha.posixid}
        supplementalGroups: [${skaha.supgroups}]
        runAsNonRoot: true
```
so none of those need to be set again in containers, while OTOH other container-level security controls need to be applied.

This MR:
- removes redundant settings in the container sec Con that were applied already in the pod sec Con
- applies additional explicit security settings in container sec Cons that can only be applied there

Overall this should make the different securityContexts more explicit and consistent and make Skaha more secure and portable.

### DIFF
--- a/deployment/helm/skaha/Chart.yaml
+++ b/deployment/helm/skaha/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.16
+version: 0.4.17
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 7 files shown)


rptaylor:
@brianmajor

at88mph (approved):

PR #666, opened 2024-07-22
Posix map init users

at88mph:
Configure init containers into one place for easy modification
Skaha writes POSIX information to cache, and is read in by init-users-groups Container


brianmajor (dismissed):
This looks like the best solution so far @at88mph.

PR #669, opened 2024-07-26
Cadc 13578

kcasteels:
These changes add node affinity rules to harbor deployments to select service nodes, as well as rules for user jobs to avoid service nodes.


at88mph (approved):

brianmajor:
This setting of p-nodes has broken the behaviour of gpu scheduling.

PR #678, opened 2024-08-06
Enable job queueing for CANFAR science platform

anujachaitanya:
We worked on [https://jira.skatelescope.org/browse/SP-4266](Enable Job queueing for CANFAR Science Platform
)

Created helm chart for kueue
Modified skaha launch for jobs accordingly
Enable feature toggle for GPU resources and limits

@abhishekghoshhh  @rob-tw

We have made the changes mentioned in earlier PR.


deployment/helm/kueue/templates/visibility/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+  {{- include "kueue.labels" . | nindent 4 }}
+  name: '{{ include "kueue.fullname" . }}-visibility-server'
+  namespace: '{{ .Release.Namespace }}'
+spec:
+  ports:
+    - name: https
+      port: 443
+      protocol: TCP
+      targetPort: 8082
```
at88mph, line 14: I'm unclear how this Visibility Server works.  This service will expose something on the default `https` port, but where does the 8082 target port come from?  Also, we typically use Traefik Endpoints for this, so is this different from that?

at88mph (changes requested):
One question regarding the Visibility Service.

at88mph:
Also, sorry if this was discussed already, but is there a Kueue Helm Chart?  If so, we can install it as a dependency or sub-chart.

anujachaitanya:
> Also, sorry if this was discussed already, but is there a Kueue Helm Chart? If so, we can install it as a dependency or sub-chart.

We looked into adding helm chart of Kueue as dependancy. But kueue doesn't have helm chart repo. So we had to add folder from their official github repo. Also we tried to add this folder as a dependancy to the base. It doesn't work as base and kueue require different namespaces. 

Reference: https://github.com/kubernetes-sigs/kueue/blob/main/charts/kueue/README.md

deployment/helm/kueue/templates/visibility/service.yaml
rob-tw: > I'm unclear how this Visibility Server works. This service will expose something on the default `https` port, but where does the 8082 target port come from? Also, we typically use Traefik Endpoints for this, so is this different from that?

The Visibility service is an API endpoint used to monitor pending workloads on demand. As of the date of this comment, it is an alpha level feature and is turned off by default, including on our development cluster. It may be enabled by setting the feature gates configuration 'VisibilityOnDemand=true'

Details are available at https://kueue.sigs.k8s.io/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand/

at88mph (approved):

brianmajor:
Has this been tested on keel-dev?

at88mph:
It has not.  The Kueue configuration is ignored until formally installed.

PR #645, opened 2024-06-28
Add option to mount CVMFS in skaha deployments

t1mk1k:
Add functionality to mount CVMFS in skaha deployments, by default this is turned off.

CVMFS will be mounted in:
- launch-carta.yaml
- launch-contributed.yaml
- launch-desktop-app.yaml
- launch-headless.yaml
- launch-notebook.yaml

and NOT in:
- launch-desktop.yaml

Changes:
- Introduce a variable to enable CVMFS (`deployment.skaha.cvmfsEnabled`)
- Add two helper functions for including the CVMFS Volume and VolumeMount in `deployment/helm/skaha/templates/_helpers.tpl`
- Adjust the control logic in `deployment/helm/skaha/templates/skaha-config-configmap.yaml` to:
  - use Helm templating logic on `deployment/helm/skaha/skaha-config/*.yaml` files
  - filter out none templating files
  - introduce scoping, so `.Files` and `.Values` are accessible in the range scope

PR #681, opened 2024-08-13
SP-4539 - Multiple mounts and types

anujachaitanya:
As part of this pull request, 
1. we have introduced templates for extra volume configs in the launch files. Also have refactored volumes and volume mounts. 
  We are as of now, targeting four type of volumes. 

  -   Persistent Volume
  -   HostPath
  -   ConfigMap
  -   Secrets

  We can introduce some more types, If needed. 

CC : @abhishekghoshhh @rob-tw


at88mph (changes requested):
Typically, the Helm definitions in the `session-volumes.yaml` and `session-volumes-mounts.yaml` would all go into the `_helpers.tpl` file.  Were the new files created to separate them?

anujachaitanya:
> Typically, the Helm definitions in the `session-volumes.yaml` and `session-volumes-mounts.yaml` would all go into the `_helpers.tpl` file. Were the new files created to separate them?

Yes. Going on we may introduce different volume types, in those cases separate files would be easier to maintain.

deployment/helm/skaha/templates/session-volumes-mounts.yaml
```diff
@@ -0,0 +1,52 @@
...
+          subPath: "cavern"
+        - mountPath: "/scratch"
+          name: scratch-dir
+          subPath: "${skaha.sessionid}"
+{{- end }}
+
+{{- define "skaha.session.cartaVolumeMounts" -}}
+        - mountPath: "/skaha-system"
+          name: start-carta
+{{- end }}
```
brianmajor: I don't think we need to generalize the type-specific mounts, only the common set of mounts that will go into each session (regardless of type).  So, I would suggest returning these startup script mounts back to the launch-*.yaml files.

Another reason to do that is because there are references to those scripts in the `cmd` section.  Having those in two different places creates a dependency, and thus makes it harder to understand and maintain.

deployment/helm/skaha/templates/session-volumes-mounts.yaml
```diff
@@ -0,0 +1,52 @@
...
+          {{- end }}
+          {{- if .volumeMount.readOnly }}
+          readOnly: {{ .volumeMount.readOnly }}
+          {{- end }}
+          {{- if .volumeMount.mountPropagation }}
+          mountPropagation: {{ .volumeMount.mountPropagation }}
+          {{- end }}
+        {{- end }}
+        {{- end }}
+        - mountPath: "/etc/passwd"
```
brianmajor: I consider the mounting of /etc/passwd and /etc/group a system requirement so should returned to the launch-*.yaml files.  The others (/cavern and /scratch) should stay here as deployers may want to change those.

brianmajor (changes requested):
Thanks for this PR, it's very close to being approvable.  My two inline comments are about returning the 'system' volume mount information (stuff that will break if not there) back to the launch-*.yaml files.

Additionally, there is some contention with the CVMFS PR, here:  https://github.com/opencadc/science-platform/pull/645/files
It is essentially a concrete example of an additional mount that deployers would choose, and if fact we are likely to choose.  Could you please use this as an example in your PR on how to add an additional common mount to all sessions?  I don't know if you'll be able to integrate the `enableCVMFS` flag that's in there, but if not, that's okay.  A commented out example would be sufficient.

Let me know if you have any questions.

deployment/helm/skaha/templates/session-volumes.yaml
```diff
@@ -0,0 +1,72 @@
+{{- define "skaha.session.commonVolumes" -}}
+      {{- if .Values.sessions.extraVolumes }}
```
rptaylor, line 2: Personally I don't see the benefit of defining complex logic like this.
It is much simpler to just let the user define the desired volumes and volumeMounts with a direct definition in Helm values of arbitrary YAML, and just insert that YAML, like e.g. [this](https://gitlab.cern.ch/eos/eos-charts/-/merge_requests/50/diffs?commit_id=c863db520c1f35e4f7000af443f46a046ef171fb).  [Here](https://gitlab.cern.ch/eos/eos-charts/-/merge_requests/50/diffs?commit_id=a754e8e887e318bea4ef8ef8f3f5aa5f92f59dfd) is an example of what the usage looks like.

It is not necessary (or feasible) to enumerate every possible volume mounting option or type and have an if statement and corresponding value for each one; IMHO that adds unnecessary complexity to the chart. It is better to have complex data (structured YAML in your values file) than complex code.

That being said I'm not a science-platform maintainer, this is just a suggestion.

rptaylor: If done in a fully general way like that it can be used to support any type of volume mount, so it should be possible to supersede/replace any specialized specific-purpose volume mounting config like in https://github.com/opencadc/science-platform/pull/645

brianmajor: Thanks for the review, and I see your points @rptaylor   I agree with the aim of constraining the complexity.  Perhaps one benefit of this approach is that you don't have to re-apply your mounting config to the yaml launch files after a helm upgrade.

deployment/helm/skaha/templates/session-volumes-mounts.yaml
anujachaitanya: We have moved start up scripts to launch files. Thanks!

deployment/helm/skaha/templates/session-volumes-mounts.yaml
anujachaitanya: We have moved passwd and group volumes to launch files.

anujachaitanya:
@brianmajor Regarding example:

We have added examples for extraVolumes config in the sample-local-values.yaml. We are assuming user has already created the volumes and volumeMounts.

deployment/helm/skaha/templates/session-volumes.yaml
anujachaitanya: Thanks for the review @rptaylor. We added this layer in order to avoid empty values and sanitise the input data. To avoid any failures regarding the same. User can just add their volumes via values.yaml, you can refer the example in sample-local-volumes.yaml.

brianmajor:
Thanks for the updates @anujachaitanya, looking better.  

We still need to figure out how to reconcile this with PR #645  After thinking about this some more, I feel that the CVMFS PR is an important feature and it's attractive to have it all enabled with the boolean in the helm values.  So, others please chime in here, but I'm thinking that we merge #645 first and then see what the conflicts look like in this PR.  Thoughts?

deployment/helm/skaha/templates/session-volumes.yaml
rptaylor: What's wrong with empty values?
Personally IMHO the k8s API is the best place to validate and sanitize input, rather than duplicating each field of the API in the Helm chart.  Further if statements would need to be added to support cephfs, emptydirs, CSI volume types, various other options of configmaps and secrets, etc. (`kubectl explain pod.spec.volumes.secret; kubectl explain pod.spec.volumes.configMap`)

 It is also convenient to be able to just copy/paste the same format of YAML that is equally valid in a pod spec as in the helm values.  Anyway that is just my 2c.

rptaylor:
@brianmajor  I would lean towards supporting fully generalized volumes and volumeMounts, so that the same chart functionality can be used to support mounting any arbitrary volumes in any possible way. Then it would support both CVMFS volumes (in particular, both the hostPath and CSI-based approaches) and anything else too.

Personally in my charts I lean towards the most flexible approach, supporting arbitrary YAML (like I suggested [here](https://github.com/opencadc/science-platform/pull/681#discussion_r1735374289) - simple code and complex values) but the appropriate approach depends on the perspective of the chart's users and developers. For end users of the charts it can be nicer to have easy options like e.g. `enableCVMFS: true` instead of laying out all the explicit YAML details of the required CVMFS volume, but this means the chart code requires more logic, i.e. simple values, complex code.

deployment/helm/skaha/templates/session-volumes.yaml
```diff
@@ -0,0 +1,36 @@
...
+          {{- end }}
+        {{- end }}
+      {{- end }}
+      {{- end }}
+      - name: cavern-volume
+        persistentVolumeClaim:
+          claimName: skaha-workload-cavern-pvc
+      - name: scratch-dir
+        emptyDir: {}
+      - name: init-users-groups
```
brianmajor: Hi @anujachaitanya - Should this stanza (init-users-groups) be in the yaml files as well because it is more of a system concept rather than user-facing?

brianmajor (changes requested):

deployment/helm/skaha/templates/session-volumes.yaml
anujachaitanya: Hello @brianmajor, We have moved init-users-groups volumes to launch files.

brianmajor:
Thanks @anujachaitanya .  One last review by @at88mph please then we can merge and close this.

anujachaitanya:
Hello @brianmajor, As discussed yesterday. Have added the changes for SP-4265. In this PR. Please have a look at it.

deployment/helm/skaha/skaha-config/launch-carta.yaml
```diff
@@ -61,7 +61,7 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
         - /skaha-system/skaha-carta.sh ${SKAHA_TLD} ${SKAHA_TLD}/projects
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
brianmajor:
```suggestion
        imagePullPolicy: IfNotPresent
```

deployment/helm/skaha/skaha-config/launch-desktop.yaml
```diff
@@ -56,16 +56,14 @@ spec:
...
-          value: "${software.desktop.app.token}"
         securityContext:
           privileged: false
           allowPrivilegeEscalation: false
           capabilities:
             drop:
               - ALL
         image: ${software.imageid}
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
brianmajor:
```suggestion
        imagePullPolicy: IfNotPresent
```

deployment/helm/skaha/skaha-config/launch-carta.yaml
brianmajor: Pull policy should be IfNotPresent everywhere.

deployment/helm/skaha/skaha-config/launch-headless.yaml
```diff
@@ -60,7 +60,7 @@ spec:
             drop:
               - ALL
         workingDir: "${SKAHA_TLD}/home/${skaha.userid}"
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
brianmajor:
```suggestion
        imagePullPolicy: IfNotPresent
```

skaha/src/main/java/org/opencadc/skaha/image/GetAction.java
```diff
@@ -159,30 +160,36 @@ public List<Image> call() throws Exception {
...
+            throw e;
         } finally {
             if (taskExecutor != null) {
                 taskExecutor.shutdown();
             }
         }
-        
+
+        redis.put(IMAGE_LAST_UPDATED_TIMESTAMP, String.valueOf(System.currentTimeMillis()));
+        redis.put(PUBLIC_IMAGES, new ImageList(images));
```
brianmajor: How does the ImageList class an improvement to List<Image>?  If there is an advantage then perhaps this method should return `ImageList` rather than `List<Image>`?

skaha/src/main/java/org/opencadc/skaha/image/GetAction.java
```diff
@@ -102,30 +98,35 @@ public GetAction() {
...
         syncOutput.getOutputStream().write(json.getBytes());
     }
-        
+
     protected List<Image> getImages(String idToken, String typeFilter) throws Exception {
+        String lastUpdated = redis.get(IMAGE_LAST_UPDATED_TIMESTAMP);
+        if (isImageListUpdated(lastUpdated)) {
+            return redis.get(PUBLIC_IMAGES, ImageList.class).get();
+        }
+
```
brianmajor: I think there is a threading and timing issue with the cache refresh.  If you take our current production deployment as an example, the call(s) to list images from harbor takes quite a while right now, let's say 10 seconds.  During this 10 seconds, more requests to list images will come in and will also attempt to refresh the cache because the timestamp has not yet been updated.  So, one issue is that there should only be 1 thread trying to refresh the cache at any given time.

Also, by using the user's thread to do the cache refresh, it will result in an inconsistent user experience.  Some calls may use the cache and be really fast and others won't.

So it's sounding like the cache refresh should be done in the backgroud.  Maybe a k8s cron is the way to do manage the cache?  Then skaha would use the cache on each call.  However you then have a startup dependency--the cron must run once before skaha can use the cache.

Open to ideas on this, but changes are definitely needed.

brianmajor (changes requested):
I've taken a look at the recent additions related to caching the harbor image list and have made some inline comments.  However, and I could be mistaken, I had thought that the intent of that work was to remove support for proprietary images in harbor, in which case I would have expected to see the removal of imagePullSecrets in the launch*.yaml files and the removal of Authorization tokens from the calls to the harbor API.

brianmajor:
Hi @anujachaitanya - I think I had misheard what Abhishek had said during our meeting, or maybe I didn't make myself clear.  I didn't expect these last 2 commits (related to redis and harbor) to be combined with this PR.  I don't know how hard it would be to separate them again, but it would make reviewing, testing, and releasing much easier for us if they were separate.

skaha/src/main/java/org/opencadc/skaha/image/GetAction.java
abhishekghoshhh: Hi @brianmajor, there is an issue in **deserialisation** with the **ObjectMapper**, that it do not accepts something like **List<Image>.class**, we have to give it **List.class** and after deserialisation it converts into List of Map automatically. This is a problem of all classes in java which use Generics. So to avoid that we have used ImageList which is nothing but a DTO class. If List<Image> is a class member then the ObjectMapper will not have any issues to deserialise.

abhishekghoshhh:
> Hi @anujachaitanya - I think I had misheard what Abhishek had said during our meeting, or maybe I didn't make myself clear. I didn't expect these last 2 commits (related to redis and harbor) to be combined with this PR. I don't know how hard it would be to separate them again, but it would make reviewing, testing, and releasing much easier for us if they were separate.

Hi @brianmajor we might have misunderstood you in the last call. Actually we were thinking to send 2 PR separately but only yesterday we merged the redis changes. We'll revert the last merge.

abhishekghoshhh:
> I've taken a look at the recent additions related to caching the harbor image list and have made some inline comments. However, and I could be mistaken, I had thought that the intent of that work was to remove support for proprietary images in harbor, in which case I would have expected to see the removal of imagePullSecrets in the launch*.yaml files and the removal of Authorization tokens from the calls to the harbor API.

Hi @brianmajor. After discussing the acceptance criteria with Sharon, she clarified that this story is primarily focused on addressing Harbor's scalability issues. The handling of proprietary images will be dealt with separately at a later stage, which is why we haven't touched that part in this story. 
However, if it's necessary to remove the code related to proprietary images now, we can do that and push the changes. Could you please confirm whether we should proceed with removing that code?

anujachaitanya:
> > Hi @anujachaitanya - I think I had misheard what Abhishek had said during our meeting, or maybe I didn't make myself clear. I didn't expect these last 2 commits (related to redis and harbor) to be combined with this PR. I don't know how hard it would be to separate them again, but it would make reviewing, testing, and releasing much easier for us if they were separate.
> 
> Hi @brianmajor we might have misunderstood you in the last call. Actually we were thinking to send 2 PR separately but only yesterday we merged the redis changes. We'll revert the last merge.

We have reverted the changes from redis and harbor. Please have a look at it.

brianmajor:
> > > Hi @anujachaitanya - I think I had misheard what Abhishek had said during our meeting, or maybe I didn't make myself clear. I didn't expect these last 2 commits (related to redis and harbor) to be combined with this PR. I don't know how hard it would be to separate them again, but it would make reviewing, testing, and releasing much easier for us if they were separate.
> > 
> > 
> > Hi @brianmajor we might have misunderstood you in the last call. Actually we were thinking to send 2 PR separately but only yesterday we merged the redis changes. We'll revert the last merge.
> 
> We have reverted the changes from redis and harbor. Please have a look at it.

Great, thank you, we will take another look at this PR.  And yes, the redis/caching should be in a separate PR.

brianmajor:
> > I've taken a look at the recent additions related to caching the harbor image list and have made some inline comments. However, and I could be mistaken, I had thought that the intent of that work was to remove support for proprietary images in harbor, in which case I would have expected to see the removal of imagePullSecrets in the launch*.yaml files and the removal of Authorization tokens from the calls to the harbor API.
> 
> Hi @brianmajor. After discussing the acceptance criteria with Sharon, she clarified that this story is primarily focused on addressing Harbor's scalability issues. The handling of proprietary images will be dealt with separately at a later stage, which is why we haven't touched that part in this story. However, if it's necessary to remove the code related to proprietary images now, we can do that and push the changes. Could you please confirm whether we should proceed with removing that code?

Hi @abhishekghoshhh @SharonGoliath - Actually, I don't see how caching could work without removing the proprietary access to harbor.  Since each users' view of the images available for launch is specific to them (they may have read or write access to images that others don't have), the cache would have to be indexed by user, so not very useful.  The current PR has a shared cache for all users, which would be the view of the public images in harbor.  So, removing the function of proprietary access to harbor goes hand-in-hand with adding caching.  There should be two areas in which the users' harbor credentials are removed:
1. On API calls to harbor--don't use (or obtain) the idToken on those calls.
2. On job launch--don't use (or obtain) the imagePullSecret (the 'cliSecret' in harbor).

deployment/helm/skaha/skaha-config/launch-carta.yaml
```diff
@@ -79,36 +79,26 @@ spec:
           protocol: TCP
           name: frontend-port
         volumeMounts:
+        - mountPath: "/skaha-system"
+          name: start-carta
```
brianmajor, line 83: It looks like launch-carta.yaml is the only one that has the start script mount removed, and I'm thinking that was a mistake?

brianmajor (changes requested):

deployment/helm/skaha/skaha-config/launch-carta.yaml
anujachaitanya: Hello @brianmajor, We haven't removed the start-script mount. It seems the order of mounts has changed hence it might seem like its removed. 

```    
    volumeMounts:
        - mountPath: "/skaha-system"
          name: start-carta
        - mountPath: "/etc/passwd"
          name: etc-passwd
          subPath: passwd
        - mountPath: "/etc/group"
          name: etc-group
          subPath: group
        {{ template "skaha.session.commonVolumeMounts" . }}
      volumes:
      - name: etc-group
        emptyDir: { }
      - name: etc-passwd
        emptyDir: { }
      - name: start-carta
        configMap:
          name: launch-scripts
          defaultMode: 0777
      - name: init-users-groups
        configMap:
          name: init-users-groups-config
          defaultMode: 0777
      {{ template "skaha.session.commonVolumes" . }}
   ```

brianmajor: Right, sorry, my mistake!

brianmajor (approved):

at88mph (approved):

PR #684, opened 2024-08-15
Gpu config api

at88mph:
- Configurable GPU using `gpu-count:<gpu-vendor>`
  - Set the `NVIDIA_CUDA_MAJOR_VERSION` environment variable in User Sessions from querying Kubernetes 
- Cleanup to prevent needing to modify each job launch file each time using a single template
- Added lookup for `SKAHA_SERVICE_ID` environment variable locally for integration tests to run


skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -620,17 +623,24 @@ public void createSession(String type, String image, String name, Integer cores,
         jobLaunchString = setConfigValue(jobLaunchString, SOFTWARE_LIMITS_CORES, cores.toString());
         jobLaunchString = setConfigValue(jobLaunchString, SOFTWARE_LIMITS_RAM, ram + "Gi");
         jobLaunchString = setConfigValue(jobLaunchString, SOFTWARE_LIMITS_GPUS, getGPUResourceLimit(gpus));
+
+        try {
+            final int majorNVIDIACUDAVersion = CommandExecutioner.getMajorNvidiaCudaGPUVersion();
+            jobLaunchString = setConfigValue(jobLaunchString, SOFTWARE_GPU_NVIDIA_CUDA_MAJOR_VERSION, Integer.toString(majorNVIDIACUDAVersion));
```
brianmajor, line 628: Users already have access to the GPU version through their software, but this may be useful, not sure.

The general idea is to allow users to land on the right GPU (brand, version, gpu-core count).  I think the ideas from ExecutionBroker are useful here and will help us align with that potential integration. 

We currently expose at /context the content of k8s-resources.  So this is just a static config reflecting the underlying capabilities of the cluster.  Ideally, those values should come from the cluster instead.  However, that is probably beyond the scope of this story.  Also beyond the scope is adding the 'brokering' part of client interaction.

So I think, for now at least, the story is to simply let users specify, through API params, those 3 gpu conditions.  I haven't gone through this whole PR yet but I'm guessing that a lot of that is already there.  Let's chat about it tomorrow.

at88mph: Thanks.  According to [CADC-13476](https://cadc-ccda.atlassian.net/browse/CADC-13476), we wanted the Major CUDA version supplied.  This way scripts can look it up.

Also, there should be two (2) parameters specified; the `gpu-type` and the `gpus` (count) parameter.

deployment/helm/skaha/skaha-config/launch-carta.yaml
```diff
@@ -39,18 +39,7 @@ spec:
...
-        - name: skaha_username
-          value: "${skaha.userid}"
-        - name: skaha_sessionid
-          value: "${skaha.sessionid}"
-        - name: HOME
-          value: "${SKAHA_TLD}/home/${skaha.userid}"
-        - name: PWD
-          value: "${SKAHA_TLD}/home/${skaha.userid}"
-        - name: OMP_NUM_THREADS
-          value: "${software.requests.cores}"
```
brianmajor: Hi @at88mph - I'm not sure that moving these from the yaml files to the helm config is valuable, since they are specific to this very spot.  It kind of aligns with the point @rptaylor made in the multiple-mounts PR.  #681 .

at88mph: OK, done.  The `common` definition is still there as it relieves the necessity of changing every Job file each time.

at88mph:
PR is too far out of date.  Closing for now.

PR #687, opened 2024-09-05
Update for Science Portal.

at88mph:


at88mph:
Already released, so bypassing to merge.

PR #688, opened 2024-09-05
Api token fix

at88mph:
CADC-13580

- Token is no longer written to user’s home directory, but rather used in an environment variable to isolate per Desktop Session.  The Token is still needed as the Desktop calls back to the /image and /context endpoints.
- Cleaned up where it’s used, and the token is only generated for Desktop Sessions

### DIFF
--- a/deployment/helm/skaha/init-users-groups-config/init-users-groups-from-service.sh
+++ b/deployment/helm/skaha/init-users-groups-config/init-users-groups-from-service.sh
@@ -51,7 +51,7 @@ if [[ "${POSIX_MAPPER_URI}" == http* ]] ; then
     TRIMMED_POSIX_MAPPER_URL=`echo "${POSIX_MAPPER_URI}" | tr -s \/ | sed 's/\(.*\)\/$/\1/'`
     POSIX_MAPPER_CAPABILITIES_URL="${TRIMMED_POSIX_MAPPER_URL}/capabilities"
 else
-    POSIX_MAPPER_CAPABILITIES_URL=`curl -SsL ${REGISTRY_URL}/resource-caps | grep ${POSIX_MAPPER_URI} | awk -F = '{print $2}' | xargs`
+    POSIX_MAPPER_CAPABILITIES_URL=`curl -kSsL ${REGISTRY_URL}/resource-caps | grep ${POSIX_MAPPER_URI} | awk -F = '{print $2}' | xargs`
 fi
 
 echo "Using POSIX Mapper service at ${POSIX_MAPPER_CAPABILITIES_URL}"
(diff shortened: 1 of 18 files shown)


deployment/helm/skaha/init-users-groups-config/init-users-groups-from-service.sh
```diff
@@ -84,11 +84,11 @@ else
 
     # For the case of using the CADC client certificate.
     if [[ -f "${CADC_PROXY_CERT_FILE}" ]] ; then
-        curl -SsL -E ${CADC_PROXY_CERT_FILE} "${UID_URL}" | sed "s/^\([a-zA-Z0-9_\@\.\-]*\):\(.*\):::$/\1:\2::${ESCAPED_HOME}\/\1:\/sbin\/nologin/" >> "${PASSWD_FILE}"
-        curl -SsL -E ${CADC_PROXY_CERT_FILE} "${GID_URL}" >> "${GROUP_FILE}"
+        curl -kSsL -E ${CADC_PROXY_CERT_FILE} "${UID_URL}" | sed "s/^\([a-zA-Z0-9_\@\.\-]*\):\(.*\):::$/\1:\2::${ESCAPED_HOME}\/\1:\/sbin\/nologin/" >> "${PASSWD_FILE}"
```
brianmajor, line 87: Why is the -k option (disable server cert check) now here?

deployment/helm/skaha/launch-scripts/build-menu.sh
```diff
@@ -225,12 +228,12 @@ build_menu_item () {
 echo "[skaha] Start building menu."
 init
 create_merged_applications_menu
-if [ -e "${HOME}/.token/.skaha" ]; then
+if [ -z "${TOKEN}" ]; then
+  handle_error "[skaha] No credentials to call back to Skaha with."
+  exit 1
+else
   echo "token is used in build menu"
   curl_out=$(curl -s -k --header "x-auth-token-skaha:${TOKEN}" "https://${HOST}/skaha/${SKAHA_API_VERSION}/image?type=desktop-app")
```
brianmajor: If possible we should remove the -k here too.  Not if we have a mechanism for running update-ca-cert (or whatever it is) at runtime on the desktop container?

brianmajor (changes requested):
Looking about right, just a couple of comments

deployment/helm/skaha/init-users-groups-config/init-users-groups-from-service.sh
at88mph: Because injecting the CA Certificate for self-signed certificates is more work for this `initContainer`, unfortunately.  So for my testing on the `cadc.dao.nrc.ca` domain, this is necessary.  Once #666 is merged, this will all go away.

deployment/helm/skaha/launch-scripts/build-menu.sh
at88mph: This is just there as a stop-gap in case writing the API Token failed for some reason.  It can be removed though as it's likely never hit.

brianmajor (approved):

PR #691, opened 2024-09-14
Node schedule control fix

at88mph:
`affinity.nodeAffinity` will be added when appropriate, and, if `gpus` requested, will be merged with the `affinity.nodeAffinity` that is required to schedule GPU nodes.


deployment/helm/skaha/skaha-config/launch-carta.yaml
```diff
@@ -27,7 +26,7 @@ spec:
       automountServiceAccountToken: false
       enableServiceLinks: false
       restartPolicy: OnFailure
-      ${skaha.schedulegpu}
+      {{ template "skaha.job.nodeAffinity" . }}
```
brianmajor, line 29: What happens to this if the deployer doesn't add an entry to their values.yaml file?  Does skaha (at runtime) simply remove it?

deployment/helm/skaha/sample-local-values.yaml
```diff
@@ -35,6 +32,18 @@ deployment:
       minEphemeralStorage: "10Gi" # Initial request for ephemeral storage (scratch disk space)
       maxEphemeralStorage: "100Gi" # Max expansion for ephemeral storage (scratch disk space)
 
+      # When set to 'true' this flag will enable GPU node scheduling.
+      gpuEnabled: true
+
+      # Example nodeAffinity to restrict how User Sessions are scheduled.
```
brianmajor: I don't think we need a generalized scheduling of user jobs on certain nodes.  We do need however, generalized scheduling of system deployments on certain nodes.  That means that we need generalized avoidance of user jobs on those nodes.

deployment/helm/skaha/sample-local-values.yaml
```diff
@@ -35,6 +32,18 @@ deployment:
...
+      nodeAffinity:
+        # Only allow executions of User Sessions on specific Nodes.
+        requiredDuringSchedulingIgnoredDuringExecution:
+          nodeSelectorTerms:
+          - matchExpressions:
+            - key: high-compute-batch-nodes
```
brianmajor: The example here could be our system nodes (persistent nodes) to show our use case.

brianmajor (changes requested):
Just a few comments to start...

deployment/helm/skaha/skaha-config/launch-carta.yaml
at88mph: Here is the implementation of the `skaha.job.nodeAffinity` template function:
```
{{/*
The affinity for Jobs.  This will import the YAML as defined by the user in the deployment.skaha.sessions.nodeAffinity stanza.
*/}}
{{- define "skaha.job.nodeAffinity" -}}
{{- with .Values.deployment.skaha.sessions.nodeAffinity }}
      affinity:
        nodeAffinity:
{{ . | toYaml | indent 10 }}
{{- end }}
{{- end }}
```

The `with` command prevents further access if its argument is missing.  If the `nodeAffinity` is absent, it is ignored.

at88mph:
Closing to recreate to relieve stale state

PR #711, opened 2024-10-11
Private harbor fix

at88mph:
Added ability to specify `x-skaha-registry-auth` to request header in format `base64Encode(username:secret)`.
Cleanup for Checkstyle
Added to Desktop and passed to Desktop App


deployment/helm/skaha/desktop-template/start-software-sh.template
```diff
@@ -123,7 +125,7 @@ get_status() {
     if [ -z "${TOKEN}" ]; then
         handle_error "[skaha] No credentials to call back to Skaha with."
     else
-        curl_out=`curl -s -L -k --header "x-auth-token-skaha: ${TOKEN}" https://(HOST)/skaha/(SKAHA_API_VERSION)/session/${VNC_PW}/app/$1`
+        curl_out=`curl -s -L -k "${CURL_APP_SESSION_HEADERS}" https://(HOST)/skaha/(SKAHA_API_VERSION)/session/${VNC_PW}/app/$1`
```
brianmajor: I think the registry credentials are only needed for the POST calls, but I suppose it doesn't hurt to have them in here too.

deployment/helm/skaha/desktop-template/start-software-sh.template
```diff
@@ -89,7 +91,7 @@ prompt_user() {
             if [ -z "${TOKEN}" ]; then
               handle_error "[skaha] No credentials to call back to Skaha with."
             else
-              app_id=`curl -s -L -k --header "x-auth-token-skaha: ${TOKEN}" -d "image=(IMAGE_ID)" --data-urlencode "param=(NAME)" https://(HOST)/skaha/(SKAHA_API_VERSION)/session/${VNC_PW}/app`
+              app_id=`curl -s -L -k "${CURL_APP_SESSION_HEADERS}" -d "image=(IMAGE_ID)" --data-urlencode "param=(NAME)" https://(HOST)/skaha/(SKAHA_API_VERSION)/session/${VNC_PW}/app`
```
brianmajor: Don't both credentials need to be sent in?  (TOKEN and REGISTRY_AUTH)?

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Constructor to use the Base64 encoded value to obtain the credentials for an Image Registry.
+     *
+     * @param encodedValue The Base64 encoded String.  Never null.
+     * @param host  The registry host for these credentials.
+     * @return ImageRegistryAuth instance.  Never null.
+     */
+    public static ImageRegistryAuth fromEncoded(final String encodedValue, final String host) {
```
brianmajor, line 39: static?

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
```diff
@@ -0,0 +1,72 @@
...
+    public String getUsername() {
+        return username;
+    }
+
+    public byte[] getSecret() {
+        return secret;
+    }
+
+    public boolean isSecretValid() {
+        return !new String(this.secret).startsWith("\ufffd");
```
brianmajor: Is this character sequence related to the encoding?  Can the encoding be checked instead?

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -482,39 +489,35 @@ private String validateImage(String imageID, String type) throws Exception {
...
-                if (type == null) {
-                    return image.getTypes().iterator().next();
-                } else {
-                    if (image.getTypes().contains(type)) {
-                        return type;
-                    } else {
-                        throw new IllegalArgumentException("image/type mismatch: " + imageID + "/" + type);
-                    }
-                }
+        final String imageRegistryHost = getRegistryHost(imageID);
```
brianmajor, line 486: The 'harborHosts' config was acting as a list of accepted/valid registries.  I think we need to preserve that concept?

brianmajor: Or maybe the logic was just shifted to elsewhere?

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -482,39 +489,35 @@ private String validateImage(String imageID, String type) throws Exception {
...
+        log.debug("Image is located at " + imageRegistryHost);
+
+        final Image image = getPublicImage(imageID);
+        final String validatedType;
+
+        // Private images are also missing from this list.
+        if (image == null) {
+            log.debug("Image " + imageID + " missing from cache...");
+            // Absence of type is assumed to be headless
```
brianmajor: The comment is right but I don't see how that is supported in the code (type==null)

skaha/src/main/java/org/opencadc/skaha/utils/CommandExecutioner.java
```diff
@@ -102,6 +82,54 @@ public static void execute(final String[] command, final OutputStream standardOu
...
+        // harbor invalidates secrets with the Unicode replacement characters 'fffd'.
+        if (registryAuth.isSecretValid()) {
+            // create new secret
+            final String[] createCmd = new String[] {
+                "kubectl", "--namespace", K8SUtil.getWorkloadNamespace(), "create", "secret", "docker-registry",
+                secretName,
+                "--docker-server=" + registryAuth.getHost(),
+                "--docker-username=" + registryAuth.getUsername(),
+                "--docker-password=" + new String(registryAuth.getSecret())
+            };
```
brianmajor: This is nice

brianmajor (changes requested):

deployment/helm/skaha/desktop-template/start-software-sh.template
at88mph: Yes.  The `CURL_APP_SESSION_HEADERS` variable contains both:
https://github.com/opencadc/science-platform/pull/711/files/3a06b2af919698c588b162eecda1b8833fdc3361#diff-2de56bc8187f4afad440423d15d2324cccf69f03706a79b18809168df3053697R7

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
at88mph: This is leftover from some old check.  This is just preserving some old check that was in place, but I'm not sure what its origins are.

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
at88mph: Yes, in a builder/singleton pattern style since the encoded value won't be preserved, so putting into a constructor didn't seem appropriate:
```java
final ImageRegistryAuth registryAuth = ImageRegistryAuth.fromEncoded(encoded, "docker.io");
```

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
at88mph: That's still true.  This is to isolate what host the provided Image was for.  It gets parsed from the ID.

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
at88mph: Ah, true, will fix.  The type is defaulted here:
https://github.com/opencadc/science-platform/blob/3a06b2af919698c588b162eecda1b8833fdc3361/skaha/src/main/java/org/opencadc/skaha/session/PostAction.java#L207

deployment/helm/skaha/desktop-template/start-software-sh.template
at88mph: That's true and I got carried away with the copy & pasting.  I think in the interest of isolating the use of that authorization key, I'll make sure it's only used where needed.

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
brianmajor: Oh yeah.  When CLISecrets would expire in Harbor, shaka would see them starting with these characters, which basically means 'not encodable'.  This check won't be needed anymore because the users are providing them directly (copying them from harbor) so won't get these weird values that we were seeing through the Harbor API.

brianmajor:
@shinybrar Could you please take a look at this PR as well?

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param encodedValue The Base64 encoded String.  Never null.
+     * @param host  The registry host for these credentials.
+     * @return ImageRegistryAuth instance.  Never null.
+     */
+    public static ImageRegistryAuth fromEncoded(final String encodedValue, final String host) {
+        if (!StringUtil.hasText(encodedValue)) {
+            throw new IllegalArgumentException("Encoded auth username and key is required.");
+        }
+
+        final String decodedValue = new String(Base64.decode(encodedValue));
```
shinybrar: Should we explicitly use `StandardCharsets.UTF_8` when converting between byte arrays and strings?

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
```diff
@@ -0,0 +1,72 @@
...
+    public String getHost() {
+        return host;
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public byte[] getSecret() {
+        return secret;
```
shinybrar, line 69: Is the returned secret mutable or immutable. Do we have a practice to always keep as immutable values to deter against parsing anomalies?

shinybrar (commented):
Keep in mind,
- I have no Java Background
- These are big classes of code to get lost in

Comments
- We could use `StandardCharsets.UTF_8` explicitly when converting between byte arrays and strings for encoding and decoding.
- There is one edgecase which still goes unchecked when the user provides a valid registry server, valid registry credentials, but malformed image/tag names for private images. This will lead to a 200 for the user, but the job will fail to launch. After discussions with @at88mph -- this is a known edgecase, which is currently bound by the limitation of our system.
- Since we perform kubectl bash command underneath the hood, the sting stream created in `CommandExecutioner` can create an escape scenario, if the client secret has a escape character in the secret. 
- We should add unittests for the new code added

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
shinybrar: @at88mph confirmed, this secret is always immutable as described by the `final` prefix in Java.

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
at88mph: OK, removed.

skaha/src/main/java/org/opencadc/skaha/registry/ImageRegistryAuth.java
shinybrar: Here is a possible implementation,
```java
final String decodedValue = new String(Base64.decode(encodedValue), StandardCharsets.UTF_8);
```

shinybrar (dismissed):
LGTM

brianmajor:
@at88mph @shinybrar - Does the removal of the "headless" type affect the behaviour of the API at all?

PR #722, opened 2024-11-06
[CADC-13809] CodeQL Implementation

shinybrar:
feat(security): added CodeQL workflow for Java


at88mph (approved):
Making the `cron` schedule configurable will be a benefit one day, but this works.

PR #728, opened 2024-11-14
Private harbor rework

at88mph:
- Rename registryXXX to repositoryXXX to differentiate with IVOA Registry
- Add default tab labels in the Portal
- Remove extra `clientSecret` configuration (bug)
- Expose `/repository` endpoint to obtain repository hosts

### DIFF
--- a/skaha/src/main/webapp/service.yaml
+++ b/skaha/src/main/webapp/service.yaml
@@ -381,3 +381,22 @@ paths:
           description: Service busy
         default:
           description: Unexpected error
+  /v0/registry:
+    get:
+      description: |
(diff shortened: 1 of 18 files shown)


skaha/src/main/webapp/service.yaml
```diff
@@ -381,3 +381,22 @@ paths:
           description: Service busy
         default:
           description: Unexpected error
+  /v0/registry:
```
shinybrar: The API should be /v0/repository as discussed.

shinybrar (changes requested):
I see the OpenAPI docs, change has been made. This looks ready to merge.

PR #784, opened 2025-01-22
Session host bypass fix

at88mph:
This PR addresses a diagnosis need to set a separate hostname to bypass some proxies, or potentially host user sessions (workload) on a different cluster or host.

### Key Changes:
1. Hostname configuration from the `SKAHA_SESSIONS_HOSTNAME` environment variable.
2. Cleanup for static utility methods
3. Added unit tests


skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
```diff
@@ -68,16 +68,24 @@
 package org.opencadc.skaha;
 
 import java.util.List;
+import java.util.Objects;
 import org.apache.log4j.Logger;
 
 public class K8SUtil {
+    private static final String VNC_URL_TEMPLATE =
```
shinybrar: Just a thought,

- Maybe we should use a URL Building library to check for malformed urls, e.g. `okhttp3`
```Java
import okhttp3.HttpUrl;

HttpUrl url = new HttpUrl.Builder()
              .scheme("https")
              .host("www.example.com")
              .addPathSegment("path")
              .addQueryParameter("key", "value")
              .build();
System.out.println(url.toString());
```
- All the URI/URL work should probably reside in a separate component file, e.g. `URLBuilder`, rather than K8SUtil.

deployment/helm/README.md
```diff
@@ -315,6 +315,11 @@ deployment:
       maxCount: "3"  # Max number of sessions per user.
       minEphemeralStorage: "20Gi"   # The initial requested amount of ephemeral (local) storage.  Does NOT apply to Desktop sessions.
       maxEphemeralStorage: "200Gi"  # The maximum amount of ephemeral (local) storage to allow a Session to extend to.  Does NOT apply to Desktop sessions.
+      
+      # Optionally setup a separate host for User Sessions for Skaha to redirect to.  The HTTPS scheme is assumed.  Defaults to the Skaha hostname (.Values.deployment.hostname).
+      # Example:
```
shinybrar, line 325: Could this somehow also help with local install of the `science-platform` e.g., if you are running the Helm Chart of the Science Platform, you can point this to localhost?

skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
```diff
@@ -203,4 +211,42 @@ public static String getRedisPort() {
...
+    private static String getCartaAltSocketURL(final String host, final String sessionID) {
+        final String cleanSessionID = Objects.requireNonNull(sessionID);
+        final String cleanHost = Objects.requireNonNull(host);
+        return String.format(
+                K8SUtil.CARTA_ALT_SOCKET_URL_TEMPLATE, cleanHost, cleanSessionID, cleanHost, cleanSessionID);
+    }
+
+    public static String getNotebookURL(String host, String sessionID, String userid, String skahaTLD) {
+        final String cleanSessionID = Objects.requireNonNull(sessionID);
+        return String.format(
```
shinybrar: Rather than using `String.format` we should use a library, to catch malformed URLS and also cases when sometimes you have `//path//path/to/something` etc.

shinybrar (changes requested):
Major comment is to utilize a URL Formatting library, rather than using String.Format.  Also, it would be nice to add better docs, especially for a test-cluster setup

deployment/helm/README.md
at88mph: If `https://localhost` was setup properly, then it one could use `hostname: localhost`, yes.

at88mph:
Latest version now has URLBuilders to create the URLs.  Thanks!

shinybrar (approved):
Looks Clean, ready to merge.

PR #800, opened 2025-02-03
fix: correct hostnames for desktop and all areas

at88mph:
Desktop calls back to Skaha and needs to maintain that hostname
Fixed sloppy hostname setting from previous fix


brianmajor (approved):
Looks good.  As discussed, might want to change sessionsHostname to workloadHostname.

at88mph:
I'll rename that as part of technical debt after.  Thanks!

at88mph:
CADC-13990

PR #803, opened 2025-02-04
fix: re add the swagger documentation

at88mph:
Add swagger-ui for documentation.


at88mph:
CSD-4014

shinybrar (approved):
- I cannot really review the contents of `skaha/src/main/webapp/swagger-ui` and most likely it should not even be here after a future cleanup.
- Although HTML can be used in descriptions, we should consider markdown or plain text for improved readability.
- We should move away from hand curated spec files, to auto-generated specs from code annotations, e.g. using the `io.swagger.core.v3.swagger-gradle-plugin` plugin
```java
package com.example.resources;

import javax.ws.rs.GET;
import javax.ws.rs.Path;
import javax.ws.rs.Produces;
import javax.ws.rs.core.MediaType;
import io.swagger.v3.oas.annotations.Operation;
import io.swagger.v3.oas.annotations.responses.ApiResponse;

@Path("/hello")
public class HelloResource {

    @GET
    @Produces(MediaType.TEXT_PLAIN)
    @Operation(
        summary = "Say Hello",
        description = "Returns a greeting message.",
        responses = {
            @ApiResponse(responseCode = "200", description = "Successful operation")
        }
    )
    public String sayHello() {
        return "Hello, World!";
    }
}
```


For now, this is good, but most likely we should create refactor work reduce the technical overhead.

at88mph:
Agreed, thanks @shinybrar .

PR #810, opened 2025-02-11
Cascade cleanup

### LINKED ISSUE #727: Skaha leaks kubernetes resources
Skaha does not appear to clean up the services and ingressroutes that it creates for workload pods so there is an unbounded growth of orphaned resources on clusters that it runs on. This contributes to load on the API and etcd services. Left long enough (especially running at large scale) there is a risk of consuming all etcd storage and bringing down the cluster. It would be difficult to recover from this condition (we already use the [maximum](https://etcd.io/docs/v3.5/dev-guide/limit/) recommended etcd size) so in practice it would likely require destroying and rebuilding the cluster. 

Currently we have to occasionally remember to remind you :) to run a manual cleanup every few months to avoid this, which is a bit operationally fragile. Skaha should automatically clean up all resources it creates to avoid leaving orphaned resources indefinitely.

at88mph:
Fixes #727 
CADC-13987

Cleanup of `Job` child resources through the `ownerReferences` property.

`Job` objects are created, their `UID` is obtained, and added to the `IngressRoute` and `Service` objects of interactive session `Job`s.


at88mph:
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

skaha/src/main/java/org/opencadc/skaha/Job.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class Job {
+    private final String name;
+    private final String uid;
+    private final String sessionID;
+    private final SessionType sessionType;
+
+    public Job(final String name, final String uid, final String sessionID, final SessionType sessionType) {
+        this.name = name;
+        this.uid = uid;
```
shinybrar: For clarity, this could these be, `k8sName` and `k8sID`?

skaha/src/main/java/org/opencadc/skaha/SessionType.java
```diff
@@ -0,0 +1,77 @@
...
+    public static SessionType fromApplicationStringType(final String applicationStringType) {
+        for (SessionType type : SessionType.values()) {
+            if (type.applicationName.equalsIgnoreCase(applicationStringType)) {
+                return type;
+            }
+        }
+        throw new IllegalArgumentException("Invalid session type: " + applicationStringType);
+    }
+
+    public Path getIngressConfigPath() {
```
shinybrar, line 39: For clarity, maybe `K8SUtil.getUserHome` should be changed to `K8SUtil.getSkahaHome` in future design changes, especially when we inject files also into user home area.

skaha/src/main/java/org/opencadc/skaha/session/SessionServiceBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+ * </pre>
+ */
+public class SessionServiceBuilder {
+    private static final String RUN_LABEL_KEY = "run";
+    private static final String SELECTOR_KEY = "canfar-net-sessionID";
+
+    private final Job job;
+
+    public SessionServiceBuilder(final Job job) {
+        this.job = Objects.requireNonNull(job);
```
shinybrar: Raise a proper error message here, `this.job = Objects.requireNonNull(job, "Job cannot be null");`

skaha/src/main/java/org/opencadc/skaha/session/SessionServiceBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+                spec.selector(selector);
+            }
+        }
+
+        return Yaml.dump(service);
+    }
+
+    V1Service loadService() throws IOException {
+        return (V1Service)
+                Yaml.load(this.job.getSessionType().getServiceConfigPath().toFile());
```
shinybrar: Since this a I/O call, may we should throw it in a try catch, given our I/O related issues? What do you think?

skaha/src/main/java/org/opencadc/skaha/session/SessionServiceBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+                } else {
+                    spec = service.getSpec();
+                }
+
+                final Map<String, String> selector;
+                if (spec.getSelector() == null) {
+                    selector = new HashMap<>();
+                } else {
+                    selector = spec.getSelector();
+                }
```
shinybrar: According to my understanding, a lot of this code can be reduced by using `Objects.requireNonNullElseGet` in Java if I am not wrong, e.g. when defining labels, can it be modified to 
```
Map<String, String> labels = Objects.requireNonNullElseGet(metadata.getLabels(), HashMap::new);
```

skaha/src/main/java/org/opencadc/skaha/utils/CommandExecutioner.java
```diff
@@ -184,4 +186,37 @@ public static void changeOwnership(String path, int posixId, int groupId) throws
...
+        final String[] getJobCommand = CommandExecutioner.getJobCommand(jobName, K8SUtil.getWorkloadNamespace());
+        final StringBuilder output = new StringBuilder(CommandExecutioner.execute(getJobCommand));
+
+        if (output.toString().startsWith("'")) {
+            output.deleteCharAt(0);
+        }
+
+        if (output.toString().endsWith("'")) {
+            output.deleteCharAt(output.length() - 1);
+        }
```
shinybrar: Use regex here for trimming, e.g.
```
String cleanedOutput = output.toString().replaceAll("^'|'$", "");
```

skaha/src/main/java/org/opencadc/skaha/session/PostAction.java
```diff
@@ -504,36 +506,35 @@ private String validateImage(String imageID, String type) throws Exception {
...
                 }
             }
-        }
-        log.debug("active interactive sessions: " + count);
-        if (count >= maxUserSessions) {
-            throw new IllegalArgumentException("User " + posixPrincipal.username + " has reached the maximum of "
-                    + maxUserSessions + " active sessions.");
+            log.debug("active interactive sessions: " + count);
+            if (count >= maxUserSessions) {
+                throw new IllegalArgumentException("User " + posixPrincipal.username + " has reached the maximum of "
```
shinybrar, line 530: The method already receives `userid` as a parameter, so I am confused to why we are using `posixPrincipal.username` here?

shinybrar (dismissed):
- The use of `SessionType` makes the code cleaner and easier to maintain. I like the move of the logic related to session types into the SessionType enum

- Just some basic Java questions, more for my learning than for anything else.

The logic of this PR is sound, might just need a couple tweaks before merging.

skaha/src/main/java/org/opencadc/skaha/session/SessionServiceBuilder.java
at88mph: This isn't I/O going to CephFS, just from the Kubernetes ConfigMap, so should be very reliable.  Also, this throws the `IOException`, which means it will be passed up as an unrecoverable error.  If that happens, there's nothing we can do anyway.

skaha/src/main/java/org/opencadc/skaha/session/SessionServiceBuilder.java
at88mph: This is an improvement, thanks.  I'll make this change.

shinybrar (approved):
Ready to merge.

brianmajor:
@at88mph - The cleanup.sh script should be deleted too.

PR #846, opened 2025-04-08
Kueue job api

at88mph:
# Description (WIP)

Skaha needs to query `Job` objects as opposed to `Pod` objects as with Kueue, there may not be a `Pod` generated yet.  This will also provide an opportunity to remove another `kubectl` command.

# Changes
- Query for `Job` objects with the `BatchV1Api`.
- Correct the output for the `Pod` usage, if the Job is running.


at88mph:
Ugly PR, sorry.  The Java client relies on the caller to handle many `null` values.  It's almost worth creating a wrapper client around it to simplify it, but here is the first draft.

skaha/src/main/java/org/opencadc/skaha/session/SessionDAO.java
```diff
@@ -4,18 +4,28 @@
...
+import java.net.URISyntaxException;
 import java.time.Instant;
+import java.time.OffsetDateTime;
 import java.time.temporal.ChronoUnit;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
shinybrar: Avoid wildcard imports, since as it makes it difficult for non-expert developers like me to know what methods are coming from where?

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
```diff
@@ -315,10 +315,15 @@ public Session getSession(String forUserID, String sessionID) throws Exception {
         throw new ResourceNotFoundException("session " + sessionID + " not found");
     }
 
-    public List<Session> getAllSessions(String forUserID) throws Exception {
+    @Deprecated
```
shinybrar: We are adding a deprecated `_getUserSession` method? Add a detailed JavaDoc for this?

skaha/src/main/java/org/opencadc/skaha/session/SessionDAO.java
```diff
@@ -4,18 +4,28 @@
...
 import ca.nrc.cadc.util.StringUtil;
+import io.kubernetes.client.Metrics;
+import io.kubernetes.client.custom.ContainerMetrics;
+import io.kubernetes.client.custom.PodMetrics;
+import io.kubernetes.client.custom.PodMetricsList;
+import io.kubernetes.client.custom.Quantity;
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.Configuration;
+import io.kubernetes.client.openapi.apis.BatchV1Api;
+import io.kubernetes.client.openapi.models.*;
```
shinybrar: Wildcard imports should be discouraged.

skaha/src/main/java/org/opencadc/skaha/utils/CommonUtils.java
```diff
@@ -24,10 +27,46 @@ public static String encodeBase64(final byte[] encodedValue) {
...
+        final long longBytes = bytes.longValue();
+        final long absB = longBytes == Long.MIN_VALUE ? Long.MAX_VALUE : Math.abs(longBytes);
+
+        if (absB < 1024) {
+            return longBytes + " B";
+        }
+
+        long value = absB;
+        CharacterIterator ci = new StringCharacterIterator("KMGTPE");
+        for (int i = 40; i >= 0 && absB > 0xfffccccccccccccL >> i; i -= 10) {
```
shinybrar: Better docs for this magic hex bit-shifting?

skaha/src/main/java/org/opencadc/skaha/utils/CommonUtils.java
```diff
@@ -24,10 +27,46 @@ public static String encodeBase64(final byte[] encodedValue) {
...
         return serviceURIs.stream().findFirst().orElseThrow(IllegalStateException::new);
     }
+
+    /**
+     * Format to Binary SI values.
+     *
+     * @param bytes Bytes reported from Kubernetes.
+     * @return String formatted, never null.
+     */
+    public static String formatMemoryFromBytes(final BigDecimal bytes) {
```
shinybrar: Could we use `FileUtils.byteCountToDisplaySize(long size)` from `org.apache.commons.io`. 

https://commons.apache.org/proper/commons-io/javadocs/api-2.4/org/apache/commons/io/FileUtils.html#byteCountToDisplaySize(long)

shinybrar (commented):
Just some small comments for readability improvement.

skaha/src/main/java/org/opencadc/skaha/session/SessionDAO.java
at88mph: Agreed.  My IDE must have put this in, but Spotless wouldn't know how to resolve that.

skaha/src/main/java/org/opencadc/skaha/session/SessionAction.java
at88mph: Removing...

skaha/src/main/java/org/opencadc/skaha/utils/CommonUtils.java
at88mph: Made readable...

skaha/src/main/java/org/opencadc/skaha/utils/CommonUtils.java
at88mph: Yes, but this actually isn't used and I'm removing it.  Once my `Kubernetes-java` client changes are accepted, I'll use that.

PR #847, opened 2025-04-09
fix: properly read harbor hosts

at88mph:
# Description
Multiple (space delimited) Harbor hosts were not being recognized

# Change
- Correctly split on space, and trim the values

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
+++ b/skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
@@ -67,7 +67,9 @@
 
 package org.opencadc.skaha;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.log4j.Logger;
 
 public class K8SUtil {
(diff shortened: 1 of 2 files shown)


at88mph:
CADC-14208

skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
```diff
@@ -143,7 +145,11 @@ public static List<String> getHarborHosts() {
             return List.of();
         }
 
-        return List.of(rawHosts.split(","));
+        return K8SUtil.getHarborHosts(rawHosts);
+    }
+
+    static List<String> getHarborHosts(final String rawHosts) {
+        return Arrays.stream(rawHosts.split(" ")).map(String::trim).collect(Collectors.toList());
```
shinybrar, line 152: Changing the deliminator from `','` to `' '`. is this expected?

shinybrar (commented):
Does `getHarborHosts()` need to handle both spaces and commas as delimiters?

brianmajor (approved):

skaha/src/main/java/org/opencadc/skaha/K8SUtil.java
at88mph: No, just spaces as that's what's documented.

PR #900, opened 2025-07-31
fix: reduce sockets to redis cache

at88mph:


shinybrar (approved):
- As we discussed, creating a new `Jedis` connection for every cache operation is expensive. While it solves the socket leak, it will not improve the performance 😭
- The `deserialize` method in `RedisCache` catches `JsonProcessingException` and logs the error, but then returns null. Is that expected / will that cause null pointer issues down the road?

PR #905, opened 2025-08-06
skaha api v1 - remove skaha dependency on harbor labels (cadc-14490)

brianmajor:
This PR should be reviewed and then discussed before being considered for merging.  The intent is probably best seen in the changes to the skaha API docs (service.yaml).  In summary:
- changed name of the optional session launch param `type` to `interact`.  The change is meant to reflect that interactive mode is what the user is looking for.
- if param `interact` is not supplied, it means no interaction necessary, so it runs as a headless job
- the labels in harbor are not used for session launching.  They're only for making images visible on the science-portal, and for categorizing those images.
- the headless label in harbor will be removed if/when this is released
- underlying cadc story: cadc-14490
- side notes:
   - desktop-apps are not subject to the changes in the PR because they are launched a different part of the api
   - the `interact` (nee `type`) param governs 2 main aspects of session launching:
      1. The ingress setup, including port and path mapping
      2. The container launch script and associated params, which includes AAI integration with the container, more path mapping, storage and mounting, and other things specific to browser applications runnong on CANFAR.
   - `type` was not a great name for the param because:
      1. it is too general and does not convey it's actual purpose, which is to enable interactive use with a container
      2. it is a reserved word in python and other languages
   - the labels in harbor still serve the purpose of enabling features in the science-portal.  When image metadata support becomes part of the science-platform, it will serve that information and not rely the harbor-specific label concept.

Tagging @at88mph @shinybrar @sfabbro

### DIFF
--- a/.github/workflows/cd.edge.build.yml
+++ b/.github/workflows/cd.edge.build.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           install: true
       - name: Perform Container Registry Login
-        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        uses: docker/login-action@184bdaa0721073962dff0199f1fb9940f07167d1 # v3.5.0
         with:
           registry: images.opencadc.org
           username: ${{ secrets.SKAHA_REGISTRY_USERNAME }}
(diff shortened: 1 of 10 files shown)


brianmajor:
After some thought, 'access' may be a better name than 'interact'.  Thinking about when we want to open up multiple access methods to a session (like multiple https entry points, ssh entry point, etc...).

brianmajor:
Internal note: when a name for the param is decided on, the session document returned on a GET should also use the same name.

PR #910, opened 2025-08-14
fix(add-separate-endpoints-for-v1-skaha-in-capabilities): Transition …

brianmajor:
…to capabilities file with entry for each endpoint


shinybrar (approved):
LGTM

PR #933, opened 2025-09-12
fix: fix typo in field member and method name

at88mph:
# Description
There is a typo in the word "priority" in a few places.

# Changes
- Correct typo in the field name and method


shinybrar (approved):
These changes look good, however, api docs (openapi.yaml) mention “Succeeded/Error” while the code actually returns “Completed” and “Failed, ...”. This mismatch predates the PR, and is prolly a future work when we align the openapi spec json schema tightly with the codebase.

PR #943, opened 2025-09-29
Limit range feature gate

at88mph:
# Description
Add ability to read resource ranges from a `LimitRange` object from Kubernetes.

# Changes
- Put new `LimitRange` object in Kubernetes if `experimentalFeature` enabled (Feature Gate)
- Re-use `ResourceContexts` class to switch JSON reader.
- Add JSON Schema test to ensure JSON consistent
- Add `ExperimentalFeatures` class to enable future Feature Gates
- Add `ResourceContextDAO` object to communicate with the Kubernetes Cluster `LimitRange` object


at88mph:
The `ResourceContexts` class will remain as the source of the information.  A feature gate was added to use the appropriate JSON (`LimitRange` vs old `k8s-resources.json`).

PR #1082, opened 2026-05-19
Skaha Platform Metrics

shinybrar:
## Summary

CADC-15588: `GET /v1/session?view=stats` cluster totals from Metrics API, not nodes/pods.

- **MetricsDAO** seam + `HttpMetricsDAO` -> `SKAHA_METRICS_BACKEND_URL` + `/api/v1/metrics/platform`
- **Cluster** `capacity`/`allocated` from Metrics; **session** `maxCPUCores`/`maxRAM` from LimitRange (`spec.max`) or `k8s-resources.json` `defaultLimit`
- **503** fail-closed on stats only: `"Platform statistics unavailable"` / `"Session resource limits unavailable"`
- **NodeDAO** gone for stats; `PodResourceUsage` stays for session listings
- **Lazy** `HttpMetricsDAO` -> missing metrics URL no longer breaks `/v1/session`, `/{id}`, `view=interactive`
- **RAM** ceilings use memory-count formatter (`24G`/`20Gi`), not bytes overload (`0G` bug)
- **CPU** bad Metrics strings -> `0.0` cores, no `NumberFormatException`

## Test plan

- [x] `./gradlew clean check` (skaha)
- [x] `GetActionResourceStatsTest` — 503 paths, lazy DAO, literal `24G`/`20Gi`
- [ ] `PlatformStatsIntTest` vs live Skaha (optional `SKAHA_METRICS_INTTEST_MODE=fixture`)

## Deploy note

Needs co-deployed Metrics + `SKAHA_METRICS_BACKEND_URL` for stats. Other session GETs OK without it.


skaha/src/main/java/org/opencadc/skaha/metrics/MetricsDAO.java
at88mph, line 1: Java Interfaces define a contract for use in multiple implementations.  I do not see us supporting multiple implementations.  We _could_ conceivably have backwards compatible support that queries the Kubernetes API directly, but I think we've proven that's just not sustainable.  As such, do one of two things, please:
- Keep this and turn it into a concrete class (basically reuse the SkahaMetricsDAO) and remove the SkahaMetricsDAO.
- Remove this interface and just use the SkahaMetricsDAO everywhere

I prefer the first one as it's consistent with the other DAO objects.

skaha/src/main/java/org/opencadc/skaha/metrics/PodMetrics.java
at88mph, line 1: Since Skaha is not supposed to connect with the Kubernetes API anymore to gather metrics, and rely on the new internal Metrics Service, is the PodMetrics class just here in case somebody disables the Metrics Service?  I don't see a need for this class if that's the case, as we don't want to disable that service I don't think.

at88mph (changes requested):
Thank you for this.

We can reduce some of these changes, I think.  See comments.

skaha/src/main/java/org/opencadc/skaha/metrics/PodMetrics.java
shinybrar: `PodMetrics` will be added to the metrics backend at a later time, the goal is to support something similar to `PlatformMetrics` for skaha from any possible provider as a source.

at88mph (approved):

PR #1103, opened 2026-06-22
fix: fix for how session denied access is conveyed

at88mph:
Temporary fix until SecurityException is handled in the main runner class (`core/cadc-util`).

Also includes support for annotations in the `skaha` Helm Chart.


shinybrar (approved):
LGTM

PR #1108, opened 2026-06-29
Auth code fix

at88mph:
- Rework client for Metrics Server
  - Bug where configuration for Metrics client was required even though disabled
- Fix for CARTA cookie request size
- Fix for build menu if the share applications link already exists.


at88mph:
Splitting up this PR.

PR #1112, opened 2026-07-02
Realign Skaha session and service labels

shinybrar:
## Summary
- Rework Skaha session, service, and pod label handling to use a consistent label set across launch paths and job builders.
- Update Kubernetes and metrics code to read and apply the revised labels correctly.
- Refresh Helm templates, config values, and test fixtures to match the new label layout.
- Align metrics API documentation with the updated labeling behavior.

## Testing
- Updated and extended unit tests for session, job builder, DAO, labels, and pod metrics behavior.
- Verified the revised Helm test fixtures reflect the expected label assignments.
- Not run (not requested).


at88mph (changes requested):
Let's remove the `labelMigration`.  I don't think it'll be reliable given how slow it is, and it's only good for the first release and useless after that.  We'll release this when we switch clusters.

at88mph (approved):

PR #1135, opened 2026-07-17
fix(skaha): allow `MetricsDAO` without metrics backend configured

shinybrar:
## Summary
- When metricsBackend.enabled=false, Helm omits SKAHA_METRICS_BACKEND_URL; MetricsDAO no longer hard-requires it at construction so session listing still works
- Platform metrics (view=stats) fail closed with 503 when backend unset/unreachable

## Test plan
- [x] `./gradlew test` (skaha) — already run locally
- [ ] Confirm session list works with metricsBackend.enabled=false
- [ ] Confirm GET /v1/session?view=stats returns 503 / "Platform statistics unavailable" when Metrics unset
- [ ] Confirm staging.canfar.net (metrics enabled) still serves stats

Closes / refs: https://herzberg.atlassian.net/browse/CADC-15967

### DIFF
--- a/skaha/src/main/java/org/opencadc/skaha/metrics/MetricsDAO.java
+++ b/skaha/src/main/java/org/opencadc/skaha/metrics/MetricsDAO.java
@@ -5,6 +5,10 @@ import org.apache.log4j.Logger;
 /**
  * Central data access for all Skaha metrics: platform stats from the Metrics backend and session pod usage (Kubernetes
  * metrics API today; Metrics backend in future).
+ *
+ * <p>Platform Metrics is optional: when {@link PlatformMetricsDAO#SKAHA_METRICS_BACKEND_URL} is unset (Helm
+ * {@code metricsBackend.enabled=false}), session listing and pod usage still work; {@link #getPlatformMetrics()} fails
+ * closed so {@code view=stats} can return 503.
  */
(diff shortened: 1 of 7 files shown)


skaha/src/main/java/org/opencadc/skaha/metrics/MetricsDAO.java
```diff
@@ -46,16 +53,20 @@ public static MetricsDAO getDefault() {
      * Fetch the current platform metrics snapshot from the Metrics backend.
      *
      * @return platform metrics from the Metrics backend
-     * @throws Exception if the snapshot cannot be retrieved
+     * @throws Exception if the backend is not configured or the snapshot cannot be retrieved
      */
     public PlatformMetrics getPlatformMetrics() throws Exception {
+        if (platformMetricsDAO == null) {
+            PlatformMetricsDAO.requireBaseUrl(null);
```
at88mph, line 60: This looks like a quick way to throw an Exception.  It doesn't read well and can be misleading.  That should throw an `IllegalStateException` to clearly indicate this is a path that should not have been taken.
```java
throw new IllegalStateException("No platform metrics configured");
```

at88mph (changes requested):

at88mph (approved):

PR #1140, opened 2026-07-21
fix: update resources to match within containers

at88mph:
# Changes
- Ensure `initContainer` resource `request` and `limits` match


at88mph:
I think this is pretty benign, @shinybrar.  Let me know if there is risk.

shinybrar (approved):
LGTM

